PR #10, opened 2019-10-14
Export context consumer/provider pair to enable unit testing

colinhemphill:
Is there any reason this change wasn't adopted before the migration to v1.0.0?

https://github.com/optimizely/fullstack-labs/pull/42/files

We are pretty much blocked on unit testing any component that contains a render component from the Optimizely React SDK. It would also be really helpful if there was some brief documentation on how to integrate unit testing with the React SDK.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+export { OptimizelyContextConsumer, OptimizelyContextProvider } from './Context'
 export { OptimizelyProvider } from './Provider'
 export { OptimizelyFeature } from './Feature'
 export { withOptimizely, WithOptimizelyProps, WithoutOptimizelyProps } from './withOptimizely'


colinhemphill:
cc: @jamesopti

jamesopti:
@colinhemphill - Thanks for pinging me on this!

The reason pull #42 (in the labs repo) wasn't merged was because I couldn't get the context approach working correctly. I believe that the `context` option in Enzyme is meant to support the legacy context api.

In our own FE codebase, we ended up going with the [_wrappingComponent_](https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md#arguments) approach, where `getClientInstance()` is an instantiated instance of the SDK (using a mock datafile).

```js
import { OptimizelyProvider } from '@optimizely/react-sdk';
...
mount(<MyComponent />, {
  wrappingComponent: OptimizelyProvider,
  wrappingComponentProps: {
    optimizely: mockOptimizelyInstance,
  },
})
```

I threw up a branch where this is running & passing: https://github.com/optimizely/react-sdk/blob/james/mock_context_provider_example/src/Feature.spec.tsx#L95-L123

Let me know if this works for you!

PR #12, opened 2019-11-14
maintanance (OptimizelyProvider): Replace react-broadcast with the native React context API

jamesopti:
## Summary

Replace usage of `react-broadcast` with the native React Context API.

The `createContext` API appears to be identical in both cases, so this is just a simple swap with adding a type for the `OptimizelyContext` state.

More here: https://reactjs.org/docs/context.html#reactcreatecontext

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,11 +543,6 @@ License: [MIT](https://github.com/piotrwitek/utility-types/blob/master/LICENSE)
 Copyright &copy; 2010-2016 Robert Kieffer and other contributors
 License: [MIT](https://github.com/kelektiv/node-uuid/blob/master/LICENSE.md)
 
-[**warning**](https://github.com/BerkeleyTrue/warning)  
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -523,10 +523,6 @@ License: [MIT](https://github.com/taylorhakes/promise-polyfill/blob/master/LICEN
 Copyright &copy; 2013-present, Facebook, Inc.
 License: [MIT](https://github.com/facebook/prop-types/blob/master/LICENSE)
 
-[**react-broadcast**](https://github.com/ReactTraining/react-broadcast)
```
mjc1283: This PR might also remove some other dependencies that were pulled in by react-broadcast (can run `npx license-checker --production --json | jq 'map_values({ licenses, publisher, repository }) | del(.[][] | nulls)'` to check).

Issue #6, opened 2019-10-01
react hooks


husseinraoouf:
will this sdk ever support react hooks?

alexparish:
Migrating to React Hooks would have the added benefit of reducing bundle sizes as functional components could be utilised instead of classes that require transpilation for older browsers.

zcramos:
:+1:

Migrating to hooks would be really nice, but just exporting the `createContext({...})` from `src/Context.js` (rather than only exporting the consumer and provider) would allow consumers to take advantage of the `useContext` hook for a more pleasant API.

jamesopti:
Hooks for `useExperiment` and `useFeature` have been merged and released with #40 ! The entire context object is now exported as well, enabling consumers to write their own hooks.

PR #28, opened 2020-03-04
feat (Hooks): Add useFeature hook

jamesopti:
## Summary

This PR adds a React Hook called `useFeature`, enabling feature usage (with variables) via the hooks interface.

The API designed here is intended to be feature complete with the `<OptimizelyFeature>` component.

### Justification - Why add hooks?

Despite just being the new hotness in the React world, there are a couple good reasons to supporting the hooks paradigm in this SDK:

- Easy use of features/variables outside the context of JSX markup
- Allow consumers to avoid using class components (mostly in the interest of bundle size)
  - This could also be done by rewriting `<OptimizelyFeature>` as a functional component that itself leverages built in React hooks for state management

#### TODO

- [x] Update hook to return a third argument, `clientReady`, which will indicate whether or not the client is ready.
- [x] Add more tests for usage of the options and overrides

### Example Usage

_Directly from the README update:_

*arguments*
* `feature : string` Key of the feature
* `options : Object`
  * `autoUpdate : boolean` (optional) If true, this component will re-render in response to datafile or user changes. Default: `false`.
  * `timeout : number` (optional) Rendering timeout as described in the `OptimizelyProvider` section. Overrides any timeout set on the ancestor `OptimizelyProvider`. 
* `overrides : Object`
  * `overrideUserId : string` (optional) Override the userId for calls to `isFeatureEnabled` for this hook.
  * `overrideAttributes : optimizely.UserAttributes` (optional) Override the user attributes for calls to `isFeatureEnabled` for this hook.

```jsx
import { useEffect } from 'react';
import { useFeature } from '@optimizely/react-sdk';

function LoginComponent() {
  const [isEnabled, variables] = useFeature('feature1', { autoUpdate: true }, { /* (Optional) User overrides */ });
  useEffect(() => {
    document.title = isEnabled ? 'login1' : 'login2';
  }, [isEnabled]);

  return (
    <p>
      <a href={isEnabled ? "/login" : "/login2"}>
        {variables.loginText}
      </a>
    </p>
  )
}
```

### AutoUpdate

A utility was added to help manage auto updates of the feature and it's variables in `autoUpdate.ts`. Eventually, this utility can be used in a `useExperiment` hook.

Additionally, the `useFeature` and `useExperiment` hooks could then handle all the state management and auto update logic for `<OptimizelyFeature>` and `<OptimizelyExperiment>`, allowing those to just be thin wrapper components.

## Test Plan

Minimal unit tests have been added to test the base case(s) here. Additional testing of the optional parameters will be added once this overall approach is agreed upon.

_Note: Until we upgrade to React 16.9, there will be errors in the `hooks.spec.tsx` unit tests about not wrapping state setting test actions in `act()`. React 16.9 introduces async support for `act()`, which will allow us to wrap our invocation of `await client.onReady()`_ 

## Issues

Addresses #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ function FeatureComponent() {
 
 ## `useFeature` Hook
 
-A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retreive the status of a feature flag and its variables. This can be useful as an alternative to the `<OptimizelyFeature>` component or to use features & variables inside code that is not explicitly rendered.
+A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retrieve the status of a feature flag and its variables. This can be useful as an alternative to the `<OptimizelyFeature>` component or to use features & variables inside code that is not explicitly rendered.
 
 *arguments*
 * `feature : string` Key of the feature
--- a/README.md
+++ b/README.md
@@ -321,10 +321,10 @@ A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retrieve the status
 import { useEffect } from 'react';
 import { useFeature } from '@optimizely/react-sdk';
 
-function FeatureComponent() {
+function LoginComponent() {
   const [isEnabled, variables] = useFeature('feature1', { autoUpdate: true }, { /* (Optional) User overrides */ });
   useEffect(() => {
-    document.title = isEnabled ? 'title1' : 'title2';
+    document.title = isEnabled ? 'login1' : 'login2';
   }, [isEnabled]);
 
   return (
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ function FeatureComponent() {
 
 ## `useFeature` Hook
 
-A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retreive the status of a feature flag and it's variables. This can be useful as an alternative to the `<OptimizelyFeature>` component or to use features & variables inside code that is not explicitly rendered.
+A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retreive the status of a feature flag and its variables. This can be useful as an alternative to the `<OptimizelyFeature>` component or to use features & variables inside code that is not explicitly rendered.
 
 *arguments*
 * `feature : string` Key of the feature
--- a/README.md
+++ b/README.md
@@ -301,6 +301,43 @@ function FeatureComponent() {
 }
 ```
 
+
+## `useFeature` Hook
+
+A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retreive the status of a feature flag and it's variables. This can be useful as an alternative to the `<OptimizelyFeature>` component or to use features & variables inside code that is not explicitly rendered.
+
+*arguments*
+* `feature : string` Key of the feature
+* `options : Object`
+  * `autoUpdate : boolean` (optional) If true, this component will re-render in response to datafile or user changes. Default: `false`.
+  * `timeout : number` (optional) Rendering timeout as described in the `OptimizelyProvider` section. Overrides any timeout set on the ancestor `OptimizelyProvider`. 
+* `overrides : Object`
+  * `overrideUserId : string` (optional) Override the userId for calls to `isFeatureEnabled` for this hook.
+  * `overrideAttributes : optimizely.UserAttributes` (optional) Override the user attributes for calls to `isFeatureEnabled` for this hook.
+
+### Render something if feature is enabled
+
+```jsx
+import { useEffect } from 'react';
+import { useFeature } from '@optimizely/react-sdk';
+
+function FeatureComponent() {
+  const [isEnabled, variables] = useFeature('feature1', { autoUpdate: true }, { /* (Optional) User overrides */ });
+  useEffect(() => {
+    document.title = isEnabled ? 'title1' : 'title2';
+  }, [isEnabled]);
+
+  return (
+    <p>
+      <a href={isEnabled ? "/login" : "/login2"}>
+        {variables.loginText}
+      </a>
+    </p>
+  )
+}
+```
+
+
 ## `withOptimizely`
 
 Any component under the `<OptimizelyProvider>` can access the Optimizely `ReactSDKClient` via the higher-order component (HoC) `withOptimizely`.
@@ -406,7 +443,7 @@ Right now server side rendering is possible with a few caveats.
 
 1. You must download the datafile manually and pass in via the `datafile` option.  Can not use `sdkKey` to automatically download.
 
-2. Rendering of components must be completely synchronous (this is true for all server side rendering)
+2. Rendering of components must be completely synchronous (this is true for all server side rendering), thus the Optimizely SDK assumes that the optimizely client has been instantiated and fired it's `onReady` event already.
 
 ### Setting up `<OptimizelyProvider>`
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ The `ReactSDKClient` client created via `createInstance` is the programmatic API
   - [JavaScript: Update datafiles](https://docs.developers.optimizely.com/full-stack/docs/javascript-update-datafiles)
 
 *returns*
-- A `ReactSDKClient` instance.
+* A `ReactSDKClient` instance.
 
 ```jsx
 import { OptimizelyProvider, createInstance } from '@optimizely/react-sdk'
@@ -309,11 +309,21 @@ A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retrieve the status
 *arguments*
 * `feature : string` Key of the feature
 * `options : Object`
-  * `autoUpdate : boolean` (optional) If true, this component will re-render in response to datafile or user changes. Default: `false`.
-  * `timeout : number` (optional) Rendering timeout as described in the `OptimizelyProvider` section. Overrides any timeout set on the ancestor `OptimizelyProvider`. 
+  * `autoUpdate : boolean` (optional) If true, this hook will update the feature and it's variables in response to datafile or user changes. Default: `false`.
+  * `timeout : number` (optional) Client timeout as described in the `OptimizelyProvider` section. Overrides any timeout set on the ancestor `OptimizelyProvider`. 
 * `overrides : Object`
   * `overrideUserId : string` (optional) Override the userId for calls to `isFeatureEnabled` for this hook.
   * `overrideAttributes : optimizely.UserAttributes` (optional) Override the user attributes for calls to `isFeatureEnabled` for this hook.
+
+*returns*
+
+* `Array` of:
+  * `isFeatureEnabled : boolean` - The `isFeatureEnabled` value for the `feature` provided.
+  * `variables : VariableValuesObject` - The variable values for the `feature` provided
+  * `clientReady : boolean` - Whether or not the underlying `ReactSDKClient` instance is ready or not.
+  * `didTimeout : boolean` - Whether or not the underlying `ReactSDKClient` became ready within the allowed `timeout` range.
+
+  _Note: `clientReady` can be true even if `didTimeout` is also true. This indicates that the client became ready *after* the timeout period._
 
 ### Render something if feature is enabled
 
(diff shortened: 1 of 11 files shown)


src/autoUpdate.ts
```diff
@@ -0,0 +1,57 @@
...
+    callback: () => void
+  ) : () => void
+}
+
+/**
+ * Utility to setup listeners for changes to the datafile or user attributes and invoke the provided callback.
+ * Returns an unListen function
+ */
+export const setupAutoUpdateListeners : AutoUpdate = (optimizely, type, value, logger, callback) => {
+  if (optimizely === null) {
```
mjc1283: Can this condition be true? The type of the first argument doesn't include `null`.

src/autoUpdate.ts
```diff
@@ -0,0 +1,57 @@
...
+    return () => {};
+  }
+  const optimizelyNotificationId = optimizely.notificationCenter.addNotificationListener(
+    NOTIFICATION_TYPES.OPTIMIZELY_CONFIG_UPDATE,
+    () => {
+      logger.info(`OPTIMIZELY_CONFIG_UPDATE, re-evaluating ${type}="%s" for user="%s"`, value, optimizely.user.id);
```
mjc1283: Hmm, any reason to interpolate `type` with a template string, but `%s` the others?

src/hooks.ts
```diff
@@ -0,0 +1,117 @@
...
+import { useCallback, useContext, useEffect, useState } from 'react';
+import * as optimizely from '@optimizely/optimizely-sdk';
+import { getLogger } from '@optimizely/js-sdk-logging';
+
+import { setupAutoUpdateListeners } from './autoUpdate';
+import { VariableValuesObject, OnReadyResult } from './client'
+import { OptimizelyContext } from './Context';
+
+type UseFeatureState = {
+  isEnabled: Boolean,
```
mjc1283: Any reason to use capital-B `Boolean` here instead of primitive `boolean`?

src/hooks.ts
```diff
@@ -0,0 +1,117 @@
...
+  variables: VariableValuesObject,
+};
+
+type ClientReady = Boolean;
```
mjc1283: Does declaring a new type like this gain us anything? Readability/documentation perhaps? In other languages I've seen examples like this where you can leverage the type system to avoid mixing up primitive values representing different things in your domain. But I don't know if TS can be used this way or if that would be useful here.

src/hooks.ts
```diff
@@ -0,0 +1,117 @@
...
+ * auto updating those values based on underlying user or datafile changes.
+ */
+export const useFeature : UseFeature = (featureKey, options = {}, overrides = {}) => {
+  const { isServerSide, optimizely, timeout } = useContext(OptimizelyContext);
+  const finalReadyTimeout: number | undefined =
+      options.timeout !== undefined ? options.timeout : timeout;
+
+  // Helper function to return the current values for isEnabled and variables.
+  const getCurrentValues = useCallback(() => ({
+    isEnabled: optimizely ? optimizely.isFeatureEnabled(featureKey, overrides.overrideUserId, overrides.overrideAttributes) : false,
```
mjc1283: When is `optimizely` a falsy value? Is that when this hook is used and there is no ancestor `OptimizelyProvider`? If that is the reason, the hook will never work as intended I'm guessing. At the very least I think that deserves an error-level log. Is there any other conventional way for a hook to signal that it failed to be setup properly or can't be used in the current context? Maybe return `null`/`undefined`?

src/hooks.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    return { isEnabled: false, variables: {}};
+  });
+  const [clientReady, setClientReady] = useState(isServerSide ? true : false);
+
+  useEffect(() => {
+    if (!optimizely) {
+      return;
+    }
+    const logger = getLogger('useFeature');
```
mjc1283: Nit: I think we can set this once at the top of the module.

src/autoUpdate.ts
jamesopti: Good catch. Will update!

src/autoUpdate.ts
jamesopti: I copied this from `Feature.tsx` and then decided to use interpolation for the variable I added.

I'll update these to all use interpolation.

src/hooks.ts
jamesopti: No - This was my naivety - I thought TS interpreted them the same but just educated myself here!

https://stackoverflow.com/a/42654496/1171978

src/hooks.ts
jamesopti: So I did this to document specifically the return interface of `useFeature` (see `UseFeature` below).

I'm totally open to other ideas here, but since the convention of Hooks is to return a tuple like list, I wanted the types to be self documenting about what you get back.

E.g.:

```js
[
    UseFeatureState["isEnabled"],
    UseFeatureState["variables"],
    ClientReady
]
```

instead of just 

```js
[
    UseFeatureState["isEnabled"],
    UseFeatureState["variables"],
    boolean
]
```

src/hooks.ts
jamesopti: Discussed IRL that it's not valid for any of our hooks or components to be used outside of an `<OptimizelyProvider>` tree with a valid `optimizely` prop.

src/hooks.ts
mjc1283: I see the gain in readability 👍

README.md
```diff
@@ -301,6 +301,43 @@ function FeatureComponent() {
...
 
+
+## `useFeature` Hook
+
+A [React Hook](https://reactjs.org/docs/hooks-intro.html) to retrieve the status of a feature flag and its variables. This can be useful as an alternative to the `<OptimizelyFeature>` component or to use features & variables inside code that is not explicitly rendered.
+
+*arguments*
+* `feature : string` Key of the feature
+* `options : Object`
+  * `autoUpdate : boolean` (optional) If true, this component will re-render in response to datafile or user changes. Default: `false`.
```
mjc1283: Probably we should describe what this does in terms of the data exposed from the hook, not talking about components.

README.md
```diff
@@ -301,6 +301,43 @@ function FeatureComponent() {
...
+  * `timeout : number` (optional) Rendering timeout as described in the `OptimizelyProvider` section. Overrides any timeout set on the ancestor `OptimizelyProvider`.
```
mjc1283: Same thing - not sure we should talk about rendering. It is up to the user to render whatever they want, the hook is just giving data.

src/hooks.ts
```diff
@@ -0,0 +1,141 @@
...
+            if (cleanupFns.length) {
+              setData(getCurrentValues());
+            }
+          })
+        );
+      }
+    })
+    .catch(() => {
+      /* The user promise or core client promise rejected. */
+      /* Swallow or log here? */
```
mjc1283: Agree, suggest we log at error level here.

README.md
```diff
@@ -301,6 +301,43 @@ function FeatureComponent() {
...
+  return (
+    <p>
+      <a href={isEnabled ? "/login" : "/login2"}>
+        {variables.loginText}
+      </a>
+    </p>
+  )
+}
+```
+
```
mjc1283: We should document in here the extra things the hook provides: `clientReady`, `didTimeout`

src/client.ts
```diff
@@ -30,6 +30,7 @@ type OnUserUpdateHandler = (userInfo: UserContext) => void
 export type OnReadyResult = {
   success: boolean
   reason?: string
+  dataReadyPromise?: Promise<any>
```
mjc1283: Should we update the docs for `onReady` to mention that this will be in there, if it timed out?

src/hooks.spec.tsx
```diff
@@ -0,0 +1,174 @@
...
+          <MyComponent />
+        </OptimizelyProvider>,
+      );
+      resolver.resolve({ success: true });
+      await optimizelyMock.onReady();
+      component.update();
+      expect(component.text()).toBe('false');
+    });
+
+    it.skip('should respect the timeout option passed', async () => {
```
circAssimilate: Is this not working yet? Should we have an explicit test for that option passed as `false` or is implicit undefined via the test above sufficient?

src/hooks.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+import { useCallback, useContext, useEffect, useState } from 'react';
+import * as optimizely from '@optimizely/optimizely-sdk';
+import { getLogger } from '@optimizely/js-sdk-logging';
+
+import { setupAutoUpdateListeners } from './autoUpdate';
+import { VariableValuesObject, OnReadyResult } from './client'
+import { OptimizelyContext } from './Context';
+
+const useFeatureLogger = getLogger('useFeature');
```
circAssimilate: Does this just drop this string in the log prefix? Not able to find the `js-sdk-logging` repo and am unsure how it works.

src/hooks.ts
```diff
@@ -0,0 +1,141 @@
...
+  overrideAttributes?: optimizely.UserAttributes,
+}
+
+interface UseFeature {
+  (
+    featureKey: string,
+    options?: UseFeatureOptions,
+    overrides?: UseFeatureOverrides,
+  ): [
+    UseFeatureState["isEnabled"],
```
circAssimilate: Does typescript require this to be bracket notation?

src/hooks.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Note: The react client can become ready AFTER the timeout period. 
+ *       ClientReady and DidTimeout provide signals to handle this scenario.
+ */
+export const useFeature : UseFeature = (featureKey, options = {}, overrides = {}) => {
+  const { isServerSide, optimizely, timeout } = useContext(OptimizelyContext);
+  if (!optimizely) {
+    throw new Error('optimizely prop must be supplied via a parent <OptimizelyProvider>');
+  }
+  const finalReadyTimeout: number | undefined =
+      options.timeout !== undefined ? options.timeout : timeout;
```
circAssimilate: I thought `typeof options.timeout !== 'undefined'` was more standard / preferred. Has that changed / was I mistaken?

src/hooks.ts
```diff
@@ -0,0 +1,141 @@
...
+  // Helper function to return the current values for isEnabled and variables.
+  const getCurrentValues = useCallback(() => ({
+    isEnabled: optimizely.isFeatureEnabled(featureKey, overrides.overrideUserId, overrides.overrideAttributes),
+    variables: optimizely.getFeatureVariables(featureKey, overrides.overrideUserId, overrides.overrideAttributes),
+  }), [featureKey, overrides]);
+
+  // Set the initial state immediately serverSide
+  const [ data, setData ] = useState<UseFeatureState>(() => {
+    if (isServerSide) {
+      return getCurrentValues();
```
circAssimilate: This for sure won't ever be accurate when `isServerSide` is false? Wondering if we could just always return the defaults.

src/hooks.ts
```diff
@@ -0,0 +1,141 @@
...
+        });
+    })
+    .then(() => {
+      setClientReady(true);
+      setData(getCurrentValues());
+      if (options.autoUpdate) {
+        cleanupFns.push(
+          setupAutoUpdateListeners(optimizely, 'feature', featureKey, useFeatureLogger, () => {
+            if (cleanupFns.length) {
+              setData(getCurrentValues());
```
circAssimilate: Will this be passed the `autoUpdate` option more than once?

README.md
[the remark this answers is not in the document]
jamesopti: I think that's worth exploring as a future update, but for now, the goal of this PR is to implement the existing spec/interface provided by `<OptimizelyFeature />`.

I think it's important for them to be the same, and updating the component would be a breaking change.

src/hooks.spec.tsx
jamesopti: Going to update this now! Thanks for asking.

It's a little tricky with a mock client, but I think I know what I wanna do.

src/hooks.ts
jamesopti: Yup! It's used as namespace in the log manager (as well as a logging prefix for that instance).

The logger package lives inside the javascript-sdk: https://github.com/optimizely/javascript-sdk/blob/master/packages/logging/src/logger.ts

src/hooks.ts
jamesopti: Yea, since it's not a namespace, you must access the property of a type using bracket notation.

src/hooks.ts
jamesopti: This is me mirroring the logic used in `<OptimizelyFeature />`, although I'll admit in general I'd agree with you.

Here I think it's safe because in this closure (and in the component), `timeout` is defined.

src/hooks.ts
jamesopti: So you're right that for the client side initial state, this is always going to return `false, {}`. This is required because we haven't implemented a synchronous way to check the ready state of the client instance.

As such, when running client side, the initial value will be `false, {}`, then the resolution of the client
s `onReady` promise will result in the values being set (unless a timeout happens, then we wait longer - see below).

src/hooks.ts
jamesopti: Not sure I understand your question here. Can you elaborate?

src/hooks.ts
circAssimilate: Ah, another js-sdk package. Shoulda looked there.

src/hooks.ts
circAssimilate: Good to know thanks!

src/hooks.ts
circAssimilate: Alright. I think it's reasonable to make it `type of` but nbd if you wanna keep it exactly the same.

src/hooks.ts
circAssimilate: It looked like `setData` was called on initial update but I wasn't sure if the `autoUpdate` needed to be passed again. Upon taking another look I see that `setupAutoUpdateListeners` is just called anytime the user / instance is updated so nevermind.

PR #31, opened 2020-03-11
chore (ESLint): Add eslint + typescript + prettier support

jamesopti:
## Summary

Setup ESLint w/Typescript & React support

Autolint 2 files just to make sure it works.

Once the core code is approved, I'll autolint the entire repo in another PR.

### Approach

I used the [Prettier.io recommended setup](https://prettier.io/docs/en/integrating-with-linters.html#recommended-configuration) with ESLint as well as [this guide](https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project) to help integrate Typescript.

The `.prettierrc` file was copied directly over from the javascript-sdk.

A few rules (3) were turned off so that after this PR, `eslint --fix` leaves us with no errors.

We can discuss after that if we want to turn those rules back on.


.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true,
+    },
+    ecmaVersion: 2017,
```
mjc1283: Hmm, so from the TypeScript side, the syntax we are allowed to use is a function of both the TypeScript version we use, and the compiler options (like `lib`, which is currently set at es2015). How do these play together? Are we forced to specify ecmaVersion here?

I am wondering if we would be better off just introducing Prettier for formatting, and leaving static checks to TypeScript for the time being. I wonder how much overlap there is between correctness checks in ESLint and the TS compiler's responsibility.

mjc1283: Discussed offline, there shouldn't be any interference.

jamesopti: Im going to set the ecmaVersion here to 2015 to match the current tsconfig `lib`, as you noted here!

Issue #30, opened 2020-03-11
Convert <OptimizelyFeature /> to a functional component that leverages useFeature


jamesopti:
## Summary

After introducing `useFeature` in #28 , we should update `<OptimizelyFeature>` to use it under the hood for feature+variable state management.

This is also a good opportunity to convert it to a functional component as well to reduce the complexity and footprint.

jamesopti:
Completed with #36 and released in #40

PR #32, opened 2020-03-11
enhancement (<OptimizelyFeature>): Convert to functional component that uses useFeature hook

jamesopti:
## Summary

Updates `<OptimizelyFeature>` to be a functional component that 

Also adds support for userId and userAttribute overrides.

Addresses #30 

### Release

Is this a breaking change? In theory the logic should behave the same, as #28 was designed to implement the same logic as `<OptimizelyFeature>`.

Timing wise, when `isServerSide` is false, the state setting of the correct values of the feature/variables in `useFeature` happens in a second chained _thennable_, rather than the first.

I dont think this should be considered a breaking change (and thus warrant a major release), but I wanted to point it out.

### DIFF
--- a/src/Feature.spec.tsx
+++ b/src/Feature.spec.tsx
@@ -110,6 +110,30 @@ describe('<OptimizelyFeature>', () => {
       expect(component.text()).toBe('true|bar');
     });
 
+    it('should pass the values for clientReady and didTimeout', async () => {
+      const component = mount(
+        <OptimizelyProvider optimizely={optimizelyMock} timeout={200}>
+          <OptimizelyFeature feature="feature1" timeout={100}>
+            {(isEnabled, variables, clientReady, didTimeout) =>
+              `${isEnabled ? 'true' : 'false'}|${variables.foo}|${clientReady}|${didTimeout}`
+            }
+          </OptimizelyFeature>
+        </OptimizelyProvider>
+      );
+
+      // while it's waiting for onReady()
+      expect(component.text()).toBe('');
+      resolver.resolve({ success: true });
+
+      await optimizelyMock.onReady();
+
+      component.update();
+
+      expect(optimizelyMock.isFeatureEnabled).toHaveBeenCalledWith('feature1', undefined, undefined);
+      expect(optimizelyMock.getFeatureVariables).toHaveBeenCalledWith('feature1', undefined, undefined);
+      expect(component.text()).toBe('true|bar|true|false');
+    });
+
     it('should respect a locally passed timeout prop', async () => {
       const component = mount(
         <OptimizelyProvider optimizely={optimizelyMock} timeout={200}>
@@ -131,6 +155,28 @@ describe('<OptimizelyFeature>', () => {
 
       expect(optimizelyMock.isFeatureEnabled).toHaveBeenCalledWith('feature1', undefined, undefined);
       expect(optimizelyMock.getFeatureVariables).toHaveBeenCalledWith('feature1', undefined, undefined);
+      expect(component.text()).toBe('true|bar');
+    });
+
+    it('should pass the override props through', async () => {
+      const component = mount(
+        <OptimizelyProvider optimizely={optimizelyMock} timeout={200}>
+          <OptimizelyFeature feature="feature1" overrideUserId="james123" overrideAttributes={{ betaUser: true }}>
+            {(isEnabled, variables) => `${isEnabled ? 'true' : 'false'}|${variables.foo}`}
+          </OptimizelyFeature>
+        </OptimizelyProvider>
+      );
+
+      // while it's waiting for onReady()
+      expect(component.text()).toBe('');
+      resolver.resolve({ success: true });
+
+      await optimizelyMock.onReady();
+
+      component.update();
+
+      expect(optimizelyMock.isFeatureEnabled).toHaveBeenCalledWith('feature1', 'james123', { betaUser: true });
+      expect(optimizelyMock.getFeatureVariables).toHaveBeenCalledWith('feature1', 'james123', { betaUser: true });
       expect(component.text()).toBe('true|bar');
     });
 
--- a/src/Feature.spec.tsx
+++ b/src/Feature.spec.tsx
@@ -81,8 +81,8 @@ describe('<OptimizelyFeature>', () => {
 
       component.update();
 
-      expect(optimizelyMock.isFeatureEnabled).toHaveBeenCalledWith('feature1');
-      expect(optimizelyMock.getFeatureVariables).toHaveBeenCalledWith('feature1');
+      expect(optimizelyMock.isFeatureEnabled).toHaveBeenCalledWith('feature1', undefined, undefined);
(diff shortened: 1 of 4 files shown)


src/Feature.spec.tsx
```diff
@@ -107,14 +107,14 @@ describe('<OptimizelyFeature>', () => {
 
       // while it's waiting for onReady()
       expect(component.text()).toBe('')
-      resolver.resolve({ sucess: true })
+      resolver.resolve({ success: true })
```
jamesopti: Success was misspelled in this entire test file, so all the tests were actually running as though success was `false` 🤦‍♂

src/Feature.tsx
```diff
@@ -14,157 +14,37 @@
...
-        variables,
-      }
-    } else {
-      this.state = {
-        canRender: false,
-        isEnabled: false,
-        variables: {},
-      }
-    }
+const FeatureComponent = (props: FeatureProps): any => {
```
jamesopti: Im not sure what Typescript wants for the return type of this functional component.

I tried a couple things but nothing pleased the `withOptimizely` definition of `Component: React.ComponentType<P>,`

@mjc1283 any ideas for how to improve this?

src/Feature.spec.tsx
circAssimilate: 😱

src/Feature.tsx
circAssimilate: Would it be `React.FC`?

src/Feature.tsx
```diff
@@ -14,157 +14,37 @@
...
+  const { feature, timeout, autoUpdate, children, overrideUserId, overrideAttributes } = props;
```
circAssimilate: nit: Is prettier being run on this or did we not do that yet for this SDK?

src/Feature.tsx
```diff
@@ -14,157 +14,37 @@
...
-    const { isEnabled, variables, canRender } = this.state
-
-    if (!canRender) {
-      return null
-    }
-
-    return children(isEnabled, variables)
-  }
-}
+  return children(isEnabled, variables)
```
circAssimilate: Is there value in passing `clientReady` and `didTimeout` like `useFeature` has? One usecase could be the decision to show a spinner while disabled until the client is ready then to show the default case.

src/Feature.tsx
```diff
@@ -14,157 +14,37 @@
...
-      }
-    }
+const FeatureComponent = (props: FeatureProps): any => {
+  const { feature, timeout, autoUpdate, children, overrideUserId, overrideAttributes } = props;
+  const [
+    isEnabled,
+    variables,
+    clientReady,
+    didTimeout
+  ] = useFeature(feature, { timeout, autoUpdate }, { overrideUserId, overrideAttributes });
```
circAssimilate: Should we add tests to make sure the options and overrides are passed through? Also, it seems alright to pass these even if they're not defined (looking at the tests above) but I think it could be better to construct an args array and not pass any values if they don't exist. Wdyt?

src/Feature.tsx
pauloptimizely: returning `JSX.Element` instead of `React.ReactNode` for the `children` render-prop, did the trick for me locally.

```js
export interface FeatureProps extends WithOptimizelyProps {
  feature: string
  timeout?: number
  autoUpdate?: boolean
  overrideUserId?: string
  overrideAttributes?: UserAttributes
  children: (isEnabled: boolean, variables: VariableValuesObject) => JSX.Element
}

const FeatureComponent: React.FC<FeatureProps> = (props: FeatureProps) => {
...
```

I'd have to do some research on the exact differences

circAssimilate: Paul the wizard!

jamesopti: > returning `JSX.Element` instead of `React.ReactNode` for the `children` render-prop, did the trick for me locally.

I see that works, but I think it's semantically incorrect. `props.children` can return any valid react markup, for which `ReactNode` is correct (Doesn't _have_ to be JSX):

`type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;`

I confirmed that the issue is with the `children` prop's return value, specifically.

Upon searching a little bit, it looks like I'm not the only one!

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18051

The suggestion to wrap the return value in a fragment works (`<>{children(...)}</>`), but to me feels incorrect/clunky, as it's technically modifying the react tree.

What should we do? I suppose returning the children inside a fragment isnt the worst, but I might prefer to tsignore this.

jamesopti: I think my vote is to use `any` to get around this and leave a note:

`const FeatureComponent: React.FunctionComponent<FeatureProps> = (props): any => {`

src/Feature.tsx
jamesopti: I like this idea. Will add it in

src/Feature.tsx
mjc1283: @jamesopti Does the fragment workaround give better type safety than using `any`?

pauloptimizely: Using `any` works for me, until this is fixed upstream here https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18051.

If we don't want to wait and prefer better type safety we could change the withOptimizely signature

jamesopti: > @jamesopti Does the fragment workaround give better type safety than using any?

It gives better type safety, but at the expense of changing the actual compiled code (to wrap everything in a fragment).

Considering all were doing is rendering children passed by the consumer, I feel like the type safety we gain is small (the consumers `children` function returns valid react code).

But maybe that is useful? I guess injecting a Fragment into the rendered tree isnt so bad either.

src/Feature.tsx
jamesopti: > Should we add tests to make sure the options and overrides are passed through?

Done!

> Also, it seems alright to pass these even if they're not defined (looking at the tests above) but I think it could be better to construct an args array and not pass any values if they don't exist. Wdyt?

So 2 thoughts:

- Since these are positional args, you still need to account for the case where `userId` is not provided and `attributes` isnt
- `optimizely` here is the `ReactSDK` instance, which specifically checks for each argument to be `undefined`.

As such, I think this is the best approach.

src/Feature.tsx
jamesopti: I went with the Fragment. Decided its the slightly lesser of 2 evils here.

src/hooks.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { useCallback, useContext, useEffect, useState } from 'react';
-import * as optimizely from '@optimizely/optimizely-sdk';
```
jamesopti: I think this should have caused a lint error below (using the variable name optimizely when its already declared in outer scope)

@mjc1283 - Is the tsc compiler smart enough to figure out which usage is a type?

mjc1283: +1 to linting against this. In our example code we always use `optimizelySdk` as the name of the import.

PR #34, opened 2020-03-16
chore (Hooks): Move common client ready logic to a separate function

jamesopti:
## Summary

In preparation to introduce a `useExperiment` hook, this PR moves the logic for waiting for the client to become ready, setting initial state, and configuring autoupdate to a shared function, which is passed to `useEffect`

There are a lot of variables that need to be passed, but I think this is much preferable to duplicating the somewhat intricate & nuanced logic that exists there.

## Test Plan

Existing unit tests cover this functionality, and more will be added with the introduction of `useExperiment`

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -15,13 +15,18 @@
  */
 import { useCallback, useContext, useEffect, useState } from 'react';
 import { UserAttributes } from '@optimizely/optimizely-sdk';
-import { getLogger } from '@optimizely/js-sdk-logging';
+import { getLogger, LoggerFacade } from '@optimizely/js-sdk-logging';
 
 import { setupAutoUpdateListeners } from './autoUpdate';
-import { VariableValuesObject, OnReadyResult } from './client';
+import { ReactSDKClient, VariableValuesObject, OnReadyResult } from './client';
 import { OptimizelyContext } from './Context';
 
 const useFeatureLogger = getLogger('useFeature');
+
+enum HookType {
+  EXPERIMENT = 'experiment',
+  FEATURE = 'feature',
+}
 
 type UseFeatureState = {
   isEnabled: boolean;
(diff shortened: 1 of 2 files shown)


src/hooks.ts
```diff
@@ -51,66 +56,47 @@ interface UseFeature {
...
-  const [clientReady, setClientReady] = useState(isServerSide ? true : false);
-  const [didTimeout, setDidTimeout] = useState(false);
-
-  useEffect(() => {
+const initializeWhenClientReadyFn = (
+  type: HookType,
+  name: string,
+  optimizely: ReactSDKClient,
+  options: UseFeatureOptions,
+  logger: LoggerFacade,
```
mjc1283: Why do we need to pass a logger in? So far the convention seems to be that each module has its own logger. If the goal is to distinguish different callers by the log message, logging the `type` could work for that.

src/hooks.ts
```diff
@@ -51,66 +56,47 @@ interface UseFeature {
...
-
-  const [clientReady, setClientReady] = useState(isServerSide ? true : false);
-  const [didTimeout, setDidTimeout] = useState(false);
-
-  useEffect(() => {
+const initializeWhenClientReadyFn = (
+  type: HookType,
+  name: string,
+  optimizely: ReactSDKClient,
+  options: UseFeatureOptions,
```
mjc1283: Rename `UseFeatureOptions` to a more generic name?

src/hooks.ts
```diff
@@ -120,15 +106,65 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
+  // Set the initial state immediately serverSide
+  const [data, setData] = useState<UseFeatureState>(() => {
+    if (isServerSide) {
+      return getCurrentValues();
+    }
+    return { isEnabled: false, variables: {} };
+  });
+
+  const [clientReady, setClientReady] = useState(isServerSide ? true : false);
+  const [didTimeout, setDidTimeout] = useState(false);
```
mjc1283: To reduce the number of arguments going around, is it possible to collapse all this into one `useState` call returning an object with all these things? Then set different parts of it as needed?

jamesopti: I initially thought this would be a hassle since the `useState` set state method doesnt extend by default, making it necessary to return the entire state object.

However, I think you're right that it's worth it here, and it makes the typing simpler too.

Set state calls now look like this:

`setState((state: HookState) => Object.assign({}, state, { didTimeout: true }));`

jamesopti:
@mjc1283 @circAssimilate @pauloptimizely - Gonna merge this now to proceed but happy to address any feedback in follow up PRs!

PR #36, opened 2020-03-17
feat (Hooks): Add useExperiment hook and refactor <OptimizelyExperiment> to use it

jamesopti:
## Summary

Adds a `useExperiment` hook and updates `<OptimizelyExperiment>` to use it.

## Test Plan

Unit tests added for the new hook, refactored for the `<OptimizelyExperiment>` usage, and new tests added for the passing of overrides.

Addresses #30

### DIFF
--- a/src/Experiment.tsx
+++ b/src/Experiment.tsx
@@ -24,7 +24,7 @@ import { withOptimizely, WithOptimizelyProps } from './withOptimizely';
 
 export type ChildrenRenderFunction = (variableValues: VariableValuesObject) => React.ReactNode;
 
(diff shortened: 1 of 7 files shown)


src/Experiment.tsx
```diff
@@ -14,167 +14,67 @@
...
-
-    this.unregisterUserListener = optimizely.onUserUpdate(() => {
-      logger.info('User update, re-evaluating experiment="%s" for user="%s"', experiment, optimizely.user.id);
-      const variation = optimizely.activate(experiment);
-      this.setState({
-        variation,
-      });
-    });
-  }
+  let match: React.ReactElement<VariationProps> | null = null;
```
jamesopti: Github is being dramatic here. None of this was changed from the render method of the class component.

PR #42, opened 2020-03-31
fix: Remove unnecessary withOptimizely usage from Experiment and Feature components, and fix timeout log message

mjc1283:
Summary:

`Experiment` and `Feature` were refactored to use hooks internally. Now, they access the context set by `OptimizelyProvider` via `useContext`, so there's no need to use `withOptimizely` on them.

In this PR I removed `withOptimizely` from those components, and fixed a log message related to timeout so that it shows the correct timeout when a component timeout prop was used.

Test Plan:

Manually tested, existing unit tests pass

### DIFF
--- a/src/Experiment.tsx
+++ b/src/Experiment.tsx
@@ -19,7 +19,6 @@ import { UserAttributes } from '@optimizely/optimizely-sdk';
 
 import { useExperiment } from './hooks';
 import { VariationProps } from './Variation';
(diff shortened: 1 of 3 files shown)


mjc1283:
@circAssimilate `withOptimizely` is part of the public API - users can use it to get access to the SDK client instance. I don't think we should remove it in this PR. Maybe we can provide a hook as a replacement for `withOptimizely`, but I haven't thought much about it.

PR #50, opened 2020-05-20
fix: Export useExperiment hook from index

mjc1283:
Make `useExperiment` hook available as an export from the top level package module.


apSeek:
can this be released too ? please

PR #53, opened 2020-07-06
feat: Added getFeatureVariableJson, getFeatureVariable and getAllFeatureVariables

fayyazarshad:
## Summary
Implemented getFeatureVariableJson, getFeatureVariable and getAllFeatureVariables wrappers of JS SDK

Deprecated in Readme
image

## Test plan
Added unit tests for getFeatureVariableJson, getFeatureVariable and getAllFeatureVariables


src/client.ts
```diff
@@ -452,6 +477,84 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        user.attributes,
+    );
+  }
+
+  /**
+   * Returns values for all the variables attached to the given feature flag
+   * @param {string} featureKey
+   * @param {string} overrideUserId
+   * @param {optimizely.UserAttributes} [overrideAttributes]
+   * @returns {(object | null)}
```
zashraf1985, line 539: can we mention the correct `{ [variableKey: string]: unknown }` return type here too

src/client.ts
```diff
@@ -452,6 +477,84 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+   * Returns value for the given json variable attached to the given feature
+   * flag
+   * @param {string} feature
+   * @param {string} variable
+   * @param {string} [overrideUserId]
+   * @param {optimizely.UserAttributes} [overrideAttributes]
+   * @returns {(string | null)}
+   * @memberof OptimizelyReactSDKClient
+   */
+  public getFeatureVariableJson(
```
zashraf1985, line 490: the case has been changed to `getFeatureVariableJSON` in the javascript SDK as well. Can you change it here too please

src/client.ts
```diff
@@ -452,6 +477,84 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+    );
+  }
+
+  /**
+   * Returns dynamically-typed value of the variable attached to the given
+   * feature flag. Returns null if the feature key or variable key is invalid.
+   * @param {string} featureKey
+   * @param {string} variableKey
+   * @param {string} [overrideUserId]
+   * @param {optimizely.UserAttributes} [overrideAttributes]
```
zashraf1985, line 514: can you add `returns` to the doc comment here

src/client.ts
```diff
@@ -354,6 +375,10 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
         case 'double':
           variableObj[key] = this._client.getFeatureVariableDouble(featureKey, key, userId, userAttributes);
           break;
+
+        case 'json':
+          variableObj[key] = this._client.getFeatureVariableJson(featureKey, key, userId, userAttributes);
```
zashraf1985, line 380: pleae change this reference to `getFeatureVariableJSON`. This has been updated in javascript SDK

src/client.ts
```diff
@@ -452,6 +477,84 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+
+  /**
+   * Returns values for all the variables attached to the given feature flag
+   * @param {string} featureKey
+   * @param {string} overrideUserId
+   * @param {optimizely.UserAttributes} [overrideAttributes]
+   * @returns {(object | null)}
+   * @memberof OptimizelyReactSDKClient
+   */
+  getAllFeatureVariables(
```
zashraf1985, line 542: @mjc1283 Want your take on it. We are adding this as the wrapper to the corresponding javascript sdk method. There is another public function `getFeatureVariables` as part of the same client object. This is going to be confusing for the end user eventually. There are two functions doing the same thing with different name. The only difference is that `getFeatureVariables` sends separate notification for each variable whereas the new `getAllFeatureVariables` sends one single notification containing all variables. Do you want us to do anything about it as part of this PR or you want to revisit later.?

src/client.spec.ts
```diff
@@ -38,6 +38,9 @@ describe('ReactSDKClient', () => {
       getForcedVariation: jest.fn(() => null),
       getFeatureVariableBoolean: jest.fn(() => null),
       getFeatureVariableDouble: jest.fn(() => null),
+      getFeatureVariableJson: jest.fn(() => null),
+      getAllFeatureVariables: jest.fn((): { [variableKey: string]: unknown } => { return {} }),
```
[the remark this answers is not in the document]
fayyazarshad: It turns out we don't this declaration here as well

src/client.ts
mjc1283: We should deprecate `getFeatureVariables` and remove it in the next major version. The hook that uses `getFeatureVariables` should be updated to use `getAllFeatureVariables`.

In this PR I recommend the following:
- Add a deprecated comment to `getFeatureVariables`
- Add a deprecated log message at WARNING level when `getFeatureVariables` is called
- Add a deprecated note to the README documentation for `getFeatureVariables`

mjc1283: Actually, let's skip the log warning. We don't want to be logging that every time the feature hook is used.

fayyazarshad: Deprecated and added snapshot of Readme in Description

README.md
```diff
@@ -484,7 +484,7 @@ The following type definitions are used in the `ReactSDKClient` interface:
...
 - `activate(experimentKey: string, overrideUserId?: string, overrideAttributes?: UserAttributes): string | null` Activate an experiment, and return the variation for the given user.
 - `getVariation(experimentKey: string, overrideUserId?: string, overrideAttributes?: UserAttributes): string | null` Return the variation for the given experiment and user.
-- `getFeatureVariables(featureKey: string, overrideUserId?: string, overrideAttributes?: UserAttributes): VariableValuesObject`: Decide and return variable values for the given feature and user
+- `getFeatureVariables(featureKey: string, overrideUserId?: string, overrideAttributes?: UserAttributes): VariableValuesObject`: Decide and return variable values for the given feature and user <br /> <b>Warning:</b> Deprecated since 2.0.1 <br /> `getAllFeatureVariables` is added in JavaScript SDK which is similarly returning all the feature variables, but it sends only single notification of type `all-feature-variables` instead of sending for each variable. As `getFeatureVariables` was added when this functionality wasn't provided by `JavaScript SDK`, so there is no need of it now and it would be removed in next major release
```
mjc1283, line 487: I think it should be `Deprecated since 2.1.0`. This release will be `2.1.0`.

src/client.spec.ts
```diff
@@ -38,6 +38,9 @@ describe('ReactSDKClient', () => {
       getForcedVariation: jest.fn(() => null),
       getFeatureVariableBoolean: jest.fn(() => null),
       getFeatureVariableDouble: jest.fn(() => null),
+      getFeatureVariableJSON: jest.fn(() => null),
+      getAllFeatureVariables: jest.fn(() => { return {} }),
```
mjc1283, line 42: Oh, interesting that `getAllFeatureVariables` was not here before. I guess it should be added, but does it affect any test?

PR #64, opened 2020-08-20
fix(hooks): Return latest decision values on first call, and re-render in response to decision input changes

mjc1283:
## Summary

This PR addresses two issues with the `useExperiment` and `useFeature` hooks:
- Changes to the `overrides` argument do not trigger a re-render
- Even after the client instance is ready, the first call returns `false` (for feature enabled status from `useFeature`) or `null` (for variation assignment from `useExperiment`).

A boolean-returning `isReady` method is added to `ReactSDKClient`. When `isReady` returns true, `useFeature` and `useExperiment` don't wait for the fulfillment of the `onReady` promise before calling their respective decision methods.

Arguments that affect decision results (experiment/feature key + overrides) are tracked in state, so we can compare prior and current values. When values change, the decision is recomputed. I followed the pattern described [here](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops), treating the decision state as being derived from the arguments.

The effect that subscribes to updates is also changed to depend on the decision inputs. Since override user attributes are an object, I used a ref combined with a custom equality function to ensure that the dependency array receives a new object reference only when the contents of the attributes are different.

## Test Plan
- Manually tested using an example React app. Checked `useFeature` and `useExperiment` return values both before & after client became ready. Checked that `useFeature` and `useExperiment` trigger re-renders when entity key or override arguments change.
- Unit tests: updated, and added more


src/hooks.ts
```diff
@@ -57,91 +59,202 @@ interface FeatureDecisionValues {
...
-            })
-          );
-        }
       })
       .catch(() => {
-        /* The user promise or core client promise rejected. */
-        logger.error(`Error initializing client. The core client or user promise(s) rejected.`);
+        hooksLogger.error(`Error initializing client. The core client or user promise(s) rejected.`);
       });
+  }, [optimizely, finalReadyTimeout]);
```
mikeng13, line 254: Same here about subscribing to `optimizely`

src/hooks.ts
```diff
@@ -201,38 +262,60 @@ export const useExperiment: UseExperiment = (experimentKey, options = {}, overri
...
-      return { ...initialState, ...getCurrentValues() };
+  const isClientReady = isServerSide || optimizely.isReady();
+  const decisionStateAndSetter = useState<FeatureDecisionValues>(() => {
+    if (isClientReady) {
+      return getCurrentDecision();
     }
-    return initialState;
+    return { isEnabled: false, variables: {} };
   });
+  let decisionState = decisionStateAndSetter[0];
```
circAssimilate, line 286: Is it safe to mutate the state value from useState?

src/hooks.ts
```diff
@@ -201,38 +262,60 @@ export const useExperiment: UseExperiment = (experimentKey, options = {}, overri
...
-    initializeWhenClientReadyFn(HookType.FEATURE, featureKey, optimizely, options, timeout, setState, getCurrentValues),
-    [optimizely]
-  );
+  useEffect(() => {
+    if (!isClientReady || options.autoUpdate) {
+      return setupAutoUpdateListeners(optimizely, HookType.FEATURE, featureKey, hooksLogger, () => {
+        setDecisionState(getCurrentDecision());
+      });
+    }
+    return (): void => {};
```
circAssimilate, line 310: Will we have a memory leak issue on unmount? Do we need to remove auto update listeners here?

src/hooks.ts
```diff
@@ -151,46 +202,56 @@ const initializeWhenClientReadyFn = (
...
+  // https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops
+  const currentDecisionInputs: DecisionInputs = {
+    entityKey: experimentKey,
+    overrideUserId: overrides.overrideUserId,
+    overrideAttributes: overrideAttrs,
+  };
+  const [prevDecisionInputs, setPrevDecisionInputs] = useState<DecisionInputs>(currentDecisionInputs);
+  if (!areDecisionInputsEqual(prevDecisionInputs, currentDecisionInputs)) {
+    setPrevDecisionInputs(currentDecisionInputs);
+    decisionState = getCurrentDecision();
```
circAssimilate, line 239: This feels like it's unsafe - like it would update the variable but React wouldn't necessarily know. I'm guessing you're doing it to avoid a re-render? Can you point to any docs that talk about this approach? I'm seeing https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops but it doesn't look like anything strange is happening other than a conditional state update which is expected to cause one more render. Doesn't this automatically happen on the next line when `setDecisionState()` is called?

src/hooks.ts
```diff
@@ -151,46 +202,56 @@ const initializeWhenClientReadyFn = (
...
+    setDecisionState(decisionState);
```
circAssimilate, line 240: I might be missing something obvious but I can't see what `setDecisionState` does with `decisionState`. It looks like a function without any params.

src/hooks.ts
```diff
@@ -151,46 +202,56 @@ const initializeWhenClientReadyFn = (
...
-    ),
-    [optimizely]
-  );
+  useEffect(() => {
+    if (!isClientReady || options.autoUpdate) {
+      return setupAutoUpdateListeners(optimizely, HookType.EXPERIMENT, experimentKey, hooksLogger, () => {
+        setDecisionState(getCurrentDecision());
+      });
+    }
+    return (): void => {};
```
circAssimilate, line 249: Same question as below about the need to "unlisten" on unmount.

src/hooks.ts
mjc1283: I can try taking this out. I just figured since I am updating decision state, I should return the latest value that I'm updating it to. But yea, it's not done that way in the example - even though they call `setIsScrollingdown`, they don't assign that new value to `isScrollingDown` before using it.

src/hooks.ts
mjc1283: `setDecisionState` is a function, it's the 2nd argument returned from `useState` on line 219. I am trying to update the state value here.

src/hooks.ts
mjc1283: Not sure, I will try removing this.

src/hooks.ts
mjc1283: I think it's ok because if control reaches this line, no auto update listeners were added.

circAssimilate: Oh, rightttt, my misread on this. That's clear to me now.

src/hooks.ts
```diff
@@ -201,38 +232,57 @@ export const useExperiment: UseExperiment = (experimentKey, options = {}, overri
...
-    };
-    if (isServerSide) {
-      return { ...initialState, ...getCurrentValues() };
+  const isClientReady = isServerSide || optimizely.isReady();
+  const [decisionState, setDecisionState] = useState<FeatureDecisionValues>(() => {
+    if (isClientReady) {
+      return getCurrentDecision();
     }
-    return initialState;
+    return { isEnabled: false, variables: {} };
```
circAssimilate, line 254: I am guessing there is something about this implementation of `useState` that I don’t understand. Can you help me understand what the anon function does here - does it only set initial state or is it used on every `setDecisionState`. I look at the React docs but couldn't find a clear example. The one I found seemed to insinuate that the provided function would be called every time.

circAssimilate, line 254: @mjc1283, I am guessing there is something about this implementation of `useState` that I don’t understand. Can you help me understand what the anon function does here - does it only set initial state or is it used on every `setDecisionState`. I look at the React docs but couldn't find a clear example. The one I found seemed to insinuate that the provided function would be called every time.

circAssimilate: @mjc1283 relayed this to me in Slack:

>I think the answer is: when you pass a function to useState, it is just called once the first time. It’s like an initializer. It returns the initial value of the state. As [described here under “Lazy Initial State”](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state)

PR #75, opened 2020-10-07
refactor(client): In getFeatureVariables, use OptimizelyConfig instead of internal ProjectConfig

mjc1283:
## Summary

`getFeatureVariables` reaches inside the optimizely-sdk client to access its project config manager, to find what variables are associated with a given feature key. In recent versions of optimizely-sdk, it is no longer necessary to access project config manager to get this information. Features and variables are available via the public, documented `OptimizelyConfig` interface. This PR updates `getFeatureVariables` to use `OptimizelyConfig`.

## Test Plan

- Updated unit tests
- Manual testing

PR #76, opened 2020-10-07
chore: Prepare for 2.3.2 release

yavorona:
## Summary:

- Upgrade `@optimizely/optimizely-sdk` to compatible-version dependency [4.3.4](https://github.com/optimizely/javascript-sdk/releases/tag/v4.3.4)
- Fixed return type of `getAllFeatureVariables` method in ReactSDKClient

## Test:

Existing unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/optimizely-sdk": "^4.3.3",
+    "@optimizely/optimizely-sdk": "^4.3.4",
     "hoist-non-react-statics": "^3.3.0",
     "prop-types": "^15.6.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@optimizely/js-sdk-logging": "^0.1.0",
-    "@optimizely/optimizely-sdk": "4.3.1",
+    "@optimizely/optimizely-sdk": "4.3.3",
```
mjc1283, line 32: I merged #75 , so we can make this a compatible-version dependency instead of an exact dependency:
```suggestion
    "@optimizely/optimizely-sdk": "^4.3.3",
```
I think this can be done with the `yarn add` command.

mjc1283 (changes requested):

mjc1283 (approved):

PR #80, opened 2020-10-28
build: adds umd and system build targets

### LINKED ISSUE #79: build to umd (and system?) format?
Wondering if you would be open to building to umd (and system) formats?

looking at https://github.com/optimizely/react-sdk/blob/master/scripts/build.js it seems like we could just add another line like the following

```js
exec(`./node_modules/.bin/rollup -c scripts/config.js -f umd -o dist/${packageName}.umd.js`);

// and maybe system format?
exec(`./node_modules/.bin/rollup -c scripts/config.js -f system -o dist/${packageName}.system.js`);
```

shouldn't have any side effects, since the entry points will remain the same and these are just new files being built.

if you're open to it, I can make the PR as well

frehner:
Fixes #79 

I've tried to follow the conventional commit formatting, and I've signed the agreement. :)

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -40,8 +40,8 @@ exec(`./node_modules/.bin/rollup -c scripts/config.js -f cjs -o dist/${packageNa
 
 console.log("\nBuilding UMD modules...");
 
-exec(`./node_modules/.bin/rollup -c scripts/config.js -f umd -o dist/${packageName}.umd.js`);
+exec(`./node_modules/.bin/rollup -c scripts/config.js -f umd -o dist/${packageName}.umd.js --name ${umdName}`, {EXTERNALS: 'forBrowsers'});
 
 console.log("\nBuilding SystemJS modules...");
 
-exec(`./node_modules/.bin/rollup -c scripts/config.js -f system -o dist/${packageName}.system.js`);
+exec(`./node_modules/.bin/rollup -c scripts/config.js -f system -o dist/${packageName}.system.js`, {EXTERNALS: 'forBrowsers'});
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -37,3 +37,11 @@ exec(`./node_modules/.bin/rollup -c scripts/config.js -f es -o dist/${packageNam
 console.log("\nBuilding CommonJS modules...");
 
 exec(`./node_modules/.bin/rollup -c scripts/config.js -f cjs -o dist/${packageName}.js`);
+
+console.log("\nBuilding UMD modules...");
+
+exec(`./node_modules/.bin/rollup -c scripts/config.js -f umd -o dist/${packageName}.umd.js`);
+
+console.log("\nBuilding SystemJS modules...");
+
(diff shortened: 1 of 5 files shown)


scripts/config.js
```diff
@@ -15,26 +15,30 @@
...
-function getPlugins(env) {
+function getPlugins(env, externals) {
   const plugins = [
-    resolve(),
+    nodeResolve({
+      browser: externals === 'forBrowsers',
+    }),
     commonjs({
       include: /node_modules/,
-      namedExports: { '@optimizely/js-sdk-logging': ['getLogger'] },
```
frehner, line 37: the cjs+esm interop with the older versions of rollup weren't working - when I added the umd+system builds, I also had to add a bunch of stuff to the `namedExports` field for various dependencies.

however, by updating to the latest version of rollup and commonjs plugin, they now handle that interop much better automatically and we can actually just remove this field all together.

scripts/build.js
```diff
@@ -37,3 +37,11 @@ exec(`./node_modules/.bin/rollup -c scripts/config.js -f es -o dist/${packageNam
...
 
 exec(`./node_modules/.bin/rollup -c scripts/config.js -f cjs -o dist/${packageName}.js`);
+
+console.log("\nBuilding UMD modules...");
+
+exec(`./node_modules/.bin/rollup -c scripts/config.js -f umd -o dist/${packageName}.umd.js --name ${umdName}`, {EXTERNALS: 'forBrowsers'});
+
+console.log("\nBuilding SystemJS modules...");
+
+exec(`./node_modules/.bin/rollup -c scripts/config.js -f system -o dist/${packageName}.system.js`, {EXTERNALS: 'forBrowsers'});
```
mjc1283, line 47: I tested the UMD build and saw a "process is not defined" error. Looks like one of the dependencies needs `process.env.NODE_ENV` replaced - I think it's from the `react-is` package.

This change fixes it and also enables minification:

```suggestion
exec(`./node_modules/.bin/rollup -c scripts/config.js -f umd -o dist/${packageName}.umd.js --name ${umdName}`, {EXTERNALS: 'forBrowsers', BUILD_ENV: 'production' });

console.log("\nBuilding SystemJS modules...");

exec(`./node_modules/.bin/rollup -c scripts/config.js -f system -o dist/${packageName}.system.js`, {EXTERNALS: 'forBrowsers', BUILD_ENV: 'production' });
```

@frehner what do you think? Do you want minification on the UMD and SystemJS bundles?

mjc1283 (changes requested):

scripts/build.js
frehner: great catch. yeah I think it makes sense for the umd+system builds to be minified automatically as well.

frehner:
should we change the browser field in package.json to point to the umd build too?

mjc1283 (approved):
LGTM, thanks!

> should we change the browser field in package.json to point to the umd build too?

No I don't think so. But we'll update the docs to mention these new builds.

PR #81, opened 2020-10-29
fix: logOnlyEventDispatcher matches expected type and calls done callback

mjc1283:
## Summary

- Update `logOnlyEventDispatcher` to be compatible with new type definition for `EventDispatcher`, introduced in optimizely-sdk 4.3.4
- Make `logOnlyEventDispatcher` call the done callback

## Test Plan
Added unit test


yavorona (approved):
LGTM. Thanks for fixing this!

PR #91, opened 2020-12-08
fix(hooks): Always recompute decision after resolution of ready promise

mjc1283:
## Summary

Fix the following bug: when client is initialized with datafile, and no SDK key, OptimizelyExperiment and OptimizelyFeature components don't properly render their decisions.

The issue is in the implementation in hooks and the interaction with the client's ready state (via `isReady`/`onReady`).  When initialized with datafile, the client begins in a non-ready state, but immediately becomes ready on the next tick. But the hook does not recompute the decision when this occurs - it's only recomputing the decision when a `CONFIG_UPDATE` is fired. `CONFIG_UPDATE` is never fired when the instance was initialized with datafile only.

To fix, add an effect in the `useFeature`/`useExperiment` hooks that recomputes the decision after the client becomes ready.

## Test plan

- Manual testing
- Updated unit tests

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -99,64 +99,47 @@ function areDecisionInputsEqual(oldDecisionInputs: DecisionInputs, newDecisionIn
 }
 
 /**
- * Hook providing access to initialization state for the argument client (readiness, and whether or not a timeout occurred
- * based on the timeout passed in or set on the ancestor OptimizelyProvider)
+ * Subscribe to changes in initialization state of the argument client. onInitStateChange callback
+ * is called on the following events:
+ * - optimizely successfully becomes ready
+ * - timeout is reached prior to optimizely becoming ready
+ * - optimizely becomes ready after the timeout has already passed
  * @param {ReactSDKClient} optimizely
- * @param {HookOptions} [options]
- * @returns InitializationState
+ * @param {number|undefined} timeout
+ * @param {Function} onInitStateChange
  */
-function useClientInitializationState(optimizely: ReactSDKClient, options: HookOptions = {}): InitializationState {
-  const { timeout, isServerSide } = useContext(OptimizelyContext);
-  const isClientReady = optimizely.isReady() || isServerSide;
-
-  const [initializationState, setInitializationState] = useState<InitializationState>(() => ({
-    clientReady: isClientReady,
-    didTimeout: false,
-  }));
-
-  // Update initialization state when any of these occur:
-  // - Client instance changed
-  // - Timeout (obtained from provider or passed in) changed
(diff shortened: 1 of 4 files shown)


src/hooks.ts
```diff
@@ -265,24 +268,35 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
+  useEffect(() => {
+    if (!isClientReady) {
+      subscribeToInitialization(optimizely, finalReadyTimeout, initState => {
+        setState({
+          ...getCurrentDecision(),
+          ...initState,
+        });
+      });
+    }
+  }, [isClientReady, finalReadyTimeout, setState, getCurrentDecision, optimizely]);
```
circAssimilate, line 287: Are we adding `setState` here because it was throwing a exhaustive dependency lint warning otherwise? If that function reference changes wondering if we could be over-eagerly running this effect.

src/hooks.ts
```diff
@@ -265,24 +268,35 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
+    if (options.autoUpdate) {
       return setupAutoUpdateListeners(optimizely, HookType.FEATURE, featureKey, hooksLogger, () => {
-        setDecisionState(getCurrentDecision());
+        setState(prevState => ({
+          ...prevState,
+          ...getCurrentDecision(),
+        }));
       });
     }
     return (): void => {};
```
circAssimilate, line 298: more of a question on the previously existing functionality - does this do anything helpful or could it be left off? No need to change that here, jw. My understanding is it's optional and isn't helpful here.

circAssimilate: Ohh, I see - it's because we're returning the cancelation up above so probably have to always supply it.

circAssimilate (approved):
This looks good overall to me, but would love to deploy a beta and team up for a duo bash a little bit in CodeSandbox before ga deploy. What do you think? If you've extensively tested on your end then I'm comfortable with that but happy to help.

src/hooks.ts
mjc1283: Good point. The lint rule seems to allow excluding it. I will remove it.

mjc1283 (commented):
Thanks @circAssimilate, yea let's do some testing. I'll DM you.

yavorona (approved):

circAssimilate:
I was able to manually bash on this and things are looking good from my vantage point.

Here's the old SDK:

https://codesandbox.io/s/priceless-flower-2ip8d?file=/src/App.js

And here's the new one:

https://codesandbox.io/s/falling-flower-5yuwm?file=/src/App.js

Everything seems to be working better in the new.

PR #97, opened 2021-01-22
feat: useExperiment hook rerenders when client setForcedVariation is called

mjc1283:
## Summary

Add mechanism in client to subscribe to `setForcedVariation` calls. In `useExperiment`, subscribe and re-compute decision after `setForcedVariation` is called.

## Test plan
- New unit tests
- Manual testing


yavorona (approved):
Works great when tested manually. LGTM!

PR #98, opened 2021-01-26
feat(decide): Added wrapper methods for decide APIs with cleaner return type

zashraf1985:
## Summary
Added Wrapper methods for decide APIS.

## Test Plan
- Manually Tested thouroughly.
- Added unit tests.


jaeopt (approved):
LGTM

src/utils.tsx
```diff
@@ -23,6 +23,15 @@ type User = {
   attributes: optimizely.UserAttributes;
 };
 
+export type UserInfo = {
+  id: string;
+  attributes?: optimizely.UserAttributes;
+};
```
mjc1283, line 29: We already have almost the same in `client.ts`: `UserContext`. Why not use that instead of creating this?

mjc1283 (changes requested):

src/utils.tsx
zashraf1985: Its the same one. so instead of importing `UserInfo` from client.ts into utils, it made more sense to me to import something from utils in to Client. So i moved the same `UserInfo` type that you are talking about from client to utils and used it in both the places.

mjc1283 (approved):

PR #99, opened 2021-02-02
Improve types for passing user attributes

tom-sherman:
Passing `undefined` as a value in the attributes object to `setUser` results in a runtime error and experiments failing to load.

This is (kind of) stated [in the documentation](https://docs.developers.optimizely.com/full-stack/docs/pass-in-audience-attributes-react):

> You can pass strings, numbers, Booleans, and null as user attribute values

This PR properly types `setUser` so that this constraint is correctly held.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -43,7 +43,7 @@ export interface ReactSDKClient extends optimizely.Client {
   user: UserContext;
 
   onReady(opts?: { timeout?: number }): Promise<any>;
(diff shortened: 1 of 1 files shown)


mikechu-optimizely:
The `setUser`'s `userInfo` param now uses `UserInfo` type whose `attributes?` are typed as `optimizely.UserAttributes` which includes (for now  🙁) `any` as the value for the key-value pair.

PR #100, opened 2021-02-11
feat(decide): Added useDecide hook

zashraf1985:
## Summary
Added a hook which used `decide` api and returns `OptimizelyDecision` object. It also supports auto updating just like other existing hooks

## Test Plan
- Manually tested thoroughly
- Added new unit tests

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -21,7 +21,7 @@ import { getLogger, LoggerFacade } from '@optimizely/js-sdk-logging';
 import { setupAutoUpdateListeners } from './autoUpdate';
 import { ReactSDKClient, VariableValuesObject, OnReadyResult } from './client';
 import { OptimizelyContext } from './Context';
-import { areAttributesEqual, OptimizelyDecision } from './utils';
+import { areAttributesEqual, OptimizelyDecision, createFailedDecision } from './utils';
 
 const hooksLogger: LoggerFacade = getLogger('ReactSDK');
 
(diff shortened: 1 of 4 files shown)


src/hooks.ts
```diff
@@ -35,6 +35,8 @@ type HookOptions = {
   timeout?: number;
 };
 
+type decideHooksOptions = HookOptions & { decideOptions?: OptimizelyDecideOption[] };
```
jaeopt, line 38: Shouldn't it be uppercase to be consistent? "DecideHooksOptions"

src/hooks.ts
```diff
@@ -311,3 +321,79 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
+  if (!optimizely) {
+    throw new Error('optimizely prop must be supplied via a parent <OptimizelyProvider>');
+  }
+
+  const overrideAttrs = useCompareAttrsMemoize(overrides.overrideAttributes);
+  const getCurrentDecision: () => { decision: OptimizelyDecision } = useCallback(
+    () => ({
+      decision: optimizely.decide(flagKey, options.decideOptions, overrides.overrideUserId, overrideAttrs)
+    }),
+    [optimizely, flagKey, overrides.overrideUserId, overrideAttrs]
```
jaeopt, line 343: Don't we need dependency on "options.decideOptions"?
If "IGNORE_USER_PROFILE" is added to the option, it'll generate different decision.

jaeopt (changes requested):
I had a couple of questions to clarify. See my comments.

src/hooks.ts
zashraf1985: Good catch

src/hooks.ts
zashraf1985: yes, it makes sense. i fixed it.

jaeopt (approved):
LGTM

mjc1283 (approved):

Issue #117, opened 2021-07-26
type VariableValuesObject doesn't handle JSON


timothympace:
When a variable is defined as JSON in the optimizely web interface, there is no associated typing for it on [`VariableValuesObject`](https://github.com/optimizely/react-sdk/blob/57b68da7688a209f51736c1030b13d620a52ac11/src/client.ts#L23). However, the SDK handles `JSON.parse`-ing variables of this type via [`getFeatureVariableJSON`](https://github.com/optimizely/react-sdk/blob/57b68da7688a209f51736c1030b13d620a52ac11/src/client.ts#L597). This leads to a type mismtach when using `useFeature` or `<OptimizelyFeature />`.

I might just be missing a key detail here, let me know if I'm using the SDK in an invalid way.

yavorona:
@timothympace thanks for reporting the issue! We just released a new patch [v2.6.3](https://www.npmjs.com/package/@optimizely/react-sdk/v/2.6.3) with the fix.

timothympace:
@yavorona Thank you for the quick fix!

PR #118, opened 2021-07-27
fix: Update VariableValuesObject type to handle JSON type variable

yavorona:
## Summary

Update VariableValuesObject interface to handle JSON type variable.

Issue: https://github.com/optimizely/react-sdk/issues/117

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@ The following type definitions are used in the `ReactSDKClient` interface:
 
 - `UserAttributes : { [name: string]: any }`
 - `User : { id: string | null, attributes: userAttributes }`
-- `VariableValuesObject : { [key: string]: unknown }`
+- `VariableValuesObject : { [key: string]: any }`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -318,7 +318,7 @@ The following type definitions are used in the `ReactSDKClient` interface:
 
 - `UserAttributes : { [name: string]: any }`
 - `User : { id: string | null, attributes: userAttributes }`
-- `VariableValuesObject : { [key: string]: boolean | number | string | null }`
+- `VariableValuesObject : { [key: string]: unknown }`
```
zashraf1985, line 321: Not sure if we should keep it `unknown`. will the user need to cast it or provide a type now? Will it break for the users who are already using it directly as any?

zashraf1985 (changes requested):
Looks Good! but i have a concern about if this change would break for the existing users or not.. see if it makes sense to you?

zashraf1985 (approved):

PR #123, opened 2021-09-10
fix(hooks): update useFeature decisions when forced variation updates

jscarmona:
Update feature decisions when forced variations update


dmregister (approved):

jscarmona:
@yavorona Any way we can get some eyes on this?

PR #125, opened 2021-09-27
fix: Fixed redundant re-rendering issues

zashraf1985:
## Summary
Made the following re-rendering fixes.
1. When SDK is initialized using a datafile, hooks were doing an empty first render and the real decision was returned on the second render. Fixed the SDK to render the correct decision on first render when initialized synchronously.
2. When SDK is initialized with datafile and SDK key both, The hook was being evaluated three times instead of two. Fixed the redundant re-rendering in this case.

Also checked the SDK thoroughly in many combination of sync/async intialization parameters and made fixes where needed.

## Test Plan
1. All unit tests pass.
2. Tested all the hooks and components thoroughly with hard coded data file, SDK key and both along with async/sync user initialzation and autoUpdate true/false.

### DIFF
--- a/src/Experiment.spec.tsx
+++ b/src/Experiment.spec.tsx
@@ -55,6 +55,8 @@ describe('<OptimizelyExperiment>', () => {
         attributes: {},
       },
       isReady: jest.fn().mockImplementation(() => isReady),
+      getIsReadyPromiseFulfilled: () => true,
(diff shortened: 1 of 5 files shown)


yavorona (approved):
Manually thoroughly tested all components and hooks as well. LGTM!

PR #133, opened 2021-11-12
feat(ForcedDecisions): Add forced-decisions APIs to ReactSDKClient and update useDecision hook to reflect changes

cristianparcu-epi:
## Summary
Add a set of new APIs for forced-decisions to ReactSDKClient.
- setForcedDecision
- getForcedDecision
- removeForcedDecision
- removeAllForcedDecisions

Also updated `useDecision` hook to autoupdate and reflect changes when forced decisions are set and removed.

## Test Plan
- Added unit tests.
- Manually tested thoroughly.


src/client.ts
```diff
@@ -263,6 +265,34 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
   }
 
+  getUserContextInstance(userInfo: UserInfo): optimizely.OptimizelyUserContext | null {
+    let userContext: optimizely.OptimizelyUserContext | null = null;
+
+    const userId = userInfo.id || this.user.id;
+    const userAttributes = userInfo.attributes || this.user.attributes;
+
+    if (this.userContext) {
+      if (JSON.stringify(userInfo) === JSON.stringify(this.user)) {
```
zashraf1985, line 275: using `JSON.stringify` to compare objects can cause undesired results. I would prefer explicitly coparing id and attributes. You can create a separate comparison method to make the code clearner.

src/client.ts
```diff
@@ -263,6 +265,34 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        return this.userContext;
+      }
+
+      if (userId) {
+        userContext = this._client.createUserContext(userId, userAttributes);
+        return userContext;
+      }
+
+      return null;
```
zashraf1985, line 284: If you simply remove this whole part, it will be covered by next code block because both are essentially doing the exact same thing.

src/client.ts
```diff
@@ -338,21 +370,22 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+  ): OptimizelyDecision {
     const user = this.getUserContextWithOverrides(overrideUserId, overrideAttributes);
     if (user.id === null) {
       logger.info('Not Evaluating feature "%s" because userId is not set', key);
       return createFailedDecision(key, `Not Evaluating flag ${key} because userId is not set`, user);
-    }    
-    const optlyUserContext: optimizely.OptimizelyUserContext | null = this._client.createUserContext(user.id, user.attributes);
+    }
+
+    const optlyUserContext = this.getUserContextInstance(user);
```
zashraf1985, line 380: When the user provides override user and attributes to the decide call, this logic will replace the memoized user context. In my opinion.
1. Memoized userContext should only be replaced/recreated when an explicit call to `setUser` is made i.e the `user` props to `OptimizelyProvider` changes.
2. If override User Id and / or attributes are used in a call to hook or eventaully to `decide` method, it should create a temporary userContext and use that in context of that particular call.

Overriding is temporary, it should not replace the permanently memoized user passed in as prop to OptimizelyProvider

zashraf1985 (changes requested):
Great work overall!
This is a partial review. I have suggested a few changes and will suggest a few more later today. Overall, this is looking great and heading in the right direction.

src/client.ts
zashraf1985: Ignore this comment! I understood it wrong. my bad.

src/client.ts
zashraf1985: Ignore this comment, i just realized you are not overriding the memoized context in the method.

src/hooks.ts
```diff
@@ -342,23 +343,28 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
   const overrideAttrs = useCompareAttrsMemoize(overrides.overrideAttributes);
-  const getCurrentDecision: () => { decision: OptimizelyDecision } = useCallback(
-    () => ({
-      decision: optimizely.decide(flagKey, options.decideOptions, overrides.overrideUserId, overrideAttrs)
-    }),
-    [optimizely, flagKey, overrides.overrideUserId, overrideAttrs, options.decideOptions]
-  );
+  const getCurrentDecision: () => { decision: OptimizelyDecision } = () => ({
+    decision: optimizely.decide(flagKey, options.decideOptions, overrides.overrideUserId, overrideAttrs),
+  });
```
zashraf1985, line 356: This was probably done to make sure `decide` is only called if any of the inputs that it depends on, change. Removing this from a `useCallback` hook might result in it being called multiple times on same inputs

src/store.ts
```diff
@@ -0,0 +1,52 @@
+interface iState {
```
zashraf1985, line 1: This store looks great. Can you add unit tests for this whole implementation

src/hooks.ts
```diff
@@ -413,5 +428,14 @@ export const useDecision: UseDecision = (flagKey, options = {}, overrides = {})
...
   }, [optimizely.getIsReadyPromiseFulfilled(), options.autoUpdate, optimizely, flagKey, getCurrentDecision]);
 
+  useEffect(() => {
+    if (lastUserUpdate) {
+      setState(prevState => ({
+        ...prevState,
+        ...getCurrentDecision(),
+      }));
+    }
+  }, [lastUserUpdate]);
```
zashraf1985, line 438: This whole implementation for forced decision looks great. Can you add unit tests to cover all the scenarios

src/client.ts
```diff
@@ -462,6 +502,68 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     return this._client.track(eventKey, user.id, user.attributes, eventTags);
   }
 
+  /**
+   * Sets the forced decision for specified optimizely decision context.
+   * @param {optimizely.OptimizelyDecisionContext} decisionContext
+   * @param {optimizely.OptimizelyForcedDecision} forcedDecision
+   * @memberof OptimizelyReactSDKClient
+   */
+  public setForcedDecision(
```
zashraf1985, line 511: Can you add unit tests for this?

zashraf1985 (changes requested):
The Implementation looks great so far! Can you add unit tests for all the changes you made and all the new functionality you added.

src/store.ts
cristianparcu-epi: Done!

zashraf1985 (approved):

PR #135, opened 2021-11-15
avoid throwing in react hooks if the optimizely client is not available from the context

### LINKED ISSUE #134: createInstance can throw: how to graceful fallback
There are a few issues that make it impossible for the application to still work if optimizely fails to initialize:

1. `optimizely.createInstance` from `optimizely-sdk` can return `null`
https://github.com/optimizely/javascript-sdk/blob/v4.7.0/packages/optimizely-sdk/lib/index.browser.ts#L136
2. `createInstance` in the `react-sdk` does not handle the `null` case and then just throws in the constructor when accessing `this._client.onReady`
https://github.com/optimizely/react-sdk/blob/2.7.0/src/client.ts#L195-L201
3. if we catch this thrown error in our application code, we cannot render `OptimizelyProvider` because it needs an `OptimizelyReactSDKClient` instance
https://github.com/optimizely/react-sdk/blob/2.7.0/src/Provider.tsx#L27
4. conditionally rendering the `OptimizelyProvider` is not an option because the optimizely hooks from `react-sdk` throw when there is no `optimizely` instance provided via the `OptimizelyProvider`
https://github.com/optimizely/react-sdk/blob/2.7.0/src/hooks.ts#L332-L336
5. the rules of hooks disallow to render hooks conditionally, so we cannot catch this issue on the application side

I have different ideas to solve this:
1. `react-sdk` allows to render `OptimizelyProvider` with `optimizely={null}` and the optimizely hooks are modified to not throw anymore when no `optimizely` instance is available. Applications will need to catch the thrown error from the initial `createInstance` call and as a fallback use `null`. The hooks then fall back to the same values they would when optimizely is not yet ready.
2. `react-sdk` does not throw in `createInstance` when `optimizely` is `null` and rather wraps all accesses to `this._client` with a `null` check.
3. `react-sdk` provides a `DummyReactClient` that implements all the functions from `ReactSDKClient` with no functionality. Applications will need to catch the thrown error from the initial `createInstance` call and as a fallback use the `DummyReactClient` and provide it to `OptimizelyProvider`.

I would vote for option 1. because it still allows applications to detect when `createInstance` fails and allows for the maximum flexibility. Also it's backwards-compatible.

fabb:
this allows an application to catch a thrown error from `createInstance` and use `null` as a fallback for the OptimizelyProvider, keeping the application working instead of crashing.

see #134 for more info

fixes #134

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -169,19 +169,16 @@ function useCompareAttrsMemoize(value: UserAttributes | undefined): UserAttribut
  */
 export const useExperiment: UseExperiment = (experimentKey, options = {}, overrides = {}) => {
   const { optimizely, isServerSide, timeout } = useContext(OptimizelyContext);
-  if (!optimizely) {
-    throw new Error('optimizely prop must be supplied via a parent <OptimizelyProvider>');
-  }
 
   const overrideAttrs = useCompareAttrsMemoize(overrides.overrideAttributes);
   const getCurrentDecision: () => ExperimentDecisionValues = useCallback(
     () => ({
-      variation: optimizely.activate(experimentKey, overrides.overrideUserId, overrideAttrs),
+      variation: optimizely ? optimizely.activate(experimentKey, overrides.overrideUserId, overrideAttrs) : null,
     }),
     [optimizely, experimentKey, overrides.overrideUserId, overrideAttrs]
   );
 
-  const isClientReady = isServerSide || optimizely.isReady();
+  const isClientReady = isServerSide || (optimizely ? optimizely.isReady() : false);
   const [state, setState] = useState<ExperimentDecisionValues & InitializationState>(() => {
     const decisionState = isClientReady ? getCurrentDecision() : { variation: null };
     return {
@@ -210,7 +207,7 @@ export const useExperiment: UseExperiment = (experimentKey, options = {}, overri
 
   const finalReadyTimeout = options.timeout !== undefined ? options.timeout : timeout;
   useEffect(() => {
-    if (!isClientReady) {
+    if (optimizely && !isClientReady) {
       subscribeToInitialization(optimizely, finalReadyTimeout, initState => {
         setState({
           ...getCurrentDecision(),
@@ -221,7 +218,7 @@ export const useExperiment: UseExperiment = (experimentKey, options = {}, overri
   }, [isClientReady, finalReadyTimeout, getCurrentDecision, optimizely]);
 
   useEffect(() => {
-    if (options.autoUpdate) {
+    if (optimizely && options.autoUpdate) {
       return setupAutoUpdateListeners(optimizely, HookType.EXPERIMENT, experimentKey, hooksLogger, () => {
         setState(prevState => ({
           ...prevState,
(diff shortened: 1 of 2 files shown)


zdivozzo:
Thank you for doing this, this seems like a really important feature to have. A lot of ad blockers block Optimizely and we don't want the app to go down if it can't load the client

zdivozzo:
@fabb Also, a workaround we found is to initialize the optimizelyClient with a shell [datafile](https://docs.developers.optimizely.com/full-stack/docs/example-datafile). Then try to create another instance using the SDK and if it succeeds, replace it. Not ideal (I like your solution better), but it works

src/hooks.ts
```diff
@@ -331,22 +325,35 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
  */
 export const useDecision: UseDecision = (flagKey, options = {}, overrides = {}) => {
   const { optimizely, isServerSide, timeout } = useContext(OptimizelyContext);
-  if (!optimizely) {
-    throw new Error('optimizely prop must be supplied via a parent <OptimizelyProvider>');
```
fabb, line 335: without this PR, using optimizely hooks without an `<OptimizelyProvider>` would throw

zashraf1985:
Thanks for contributing. The way you are trying to fix it is by assuming that `optimizely` object passed in to `Provider` can be null and you have added checks to `Provider` and hooks to make sure they handle optimizely as `null`. Having worked on this SDK, i would fix it in a different way. I would make sure that `createInstance` from React SDK always returns the ReactSDKClient object. I would add safeguards inside `ReactSDKClient` to make sure it does not break if, for some reason, `optimizely-sdk` returns a null client. We are already picking it up and will fix it over the next few days separately.

fabb:
@zashraf1985 sounds fine, that would be the 2. solution idea from https://github.com/optimizely/react-sdk/issues/134

fabb:
@zashraf1985 there is one downside to your approach for fixing the issue: testability. If the hooks would work without OptimizelyProvider, components that contain the hooks could be more easily unit tested. In my current unit tests I use a mocked ReactSDKClient with an OptimizelyProvider, but it‘s annoying to maintain because I need to implement all the functions from the interface, and it changes with updates of the lib (e.g. `getIsReadyPromiseFulfilled` was added lately).

Develliot:
I'm with @fabb,

I think we should make the SDK less brittle. Possibly hooks shouldn't throw an error and fall over I think they should return an error as described here:

https://github.com/optimizely/react-sdk/issues/152

I also think that the provider should just be a context provider with it's own  its own useOptimiselyContext hook.  If setting the client and user id were separated wit init function(s), the provider should be able to exist without client so if something goes wrong we can set it later. This will also prevent any rehydration issues with SSR and browser renders, if client fails or user id isn't ready.

opti-jnguyen:
@Develliot - thanks for the suggestion! Though the change you suggested makes sense, we'd have to change the way we currently initialize the React SDK which may result in breaking changes for other users.

I'd recommend posting the suggestion to our [feature request board here](https://experimentation.ideas.aha.io/) and we'll prioritize it based on the number of votes it gets.

Thanks,
John

PR #147, opened 2022-01-31
fix: Exported correct type from `createInstance`

### LINKED ISSUE #49: Export OptimizelyReactSDKClient as interface or type
The return value of `createInstance` (`OptimizelyReactSDKClient`) should be exported as a type, lest we need to use `ReturnType<typeof createInstance>`

zashraf1985:
## Summary
Return type of `createInstance` was `OptimizelyReactSDKClient` which is the implementation class. Changed it to the `ReactSDKClient` interface instead.

## Test Plan
- Manually tested thoroughly.
- All existing unit tests pass.

## Issues
Probably fixed #49


zashraf1985:
@shaharyar123 I am closing this. please open it with your credentials so that i can add review feedback

PR #148, opened 2022-02-01
refactor: Return `ReactSDKClient` interface from `createInstance` instead of implementation class

### LINKED ISSUE #49: Export OptimizelyReactSDKClient as interface or type

shaharyar123:
### Summary
Return type of createInstance was OptimizelyReactSDKClient which is the implementation class. Changed it to the ReactSDKClient interface instead.

### Test Plan

- Manually tested thoroughly.
- All existing unit tests pass.

### Issues
#49

### DIFF
--- a/src/client.spec.ts
+++ b/src/client.spec.ts
@@ -18,6 +18,10 @@ jest.mock('@optimizely/optimizely-sdk');
 import * as optimizely from '@optimizely/optimizely-sdk';
 
 import { createInstance, OnReadyResult, ReactSDKClient } from './client';
+interface MockedReactSDKClient extends ReactSDKClient {
+  client: optimizely.Client;
+  initialConfig: optimizely.Config;
+}
 
 describe('ReactSDKClient', () => {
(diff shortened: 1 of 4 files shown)


src/client.spec.ts
```diff
@@ -18,6 +18,10 @@ jest.mock('@optimizely/optimizely-sdk');
 import * as optimizely from '@optimizely/optimizely-sdk';
 
 import { createInstance, OnReadyResult, ReactSDKClient } from './client';
+interface MockedReactSDKClient extends ReactSDKClient {
+  client: optimizely.Client;
+  initialConfig: optimizely.Config;
+}
```
zashraf1985, line 24: This is clever. Great Job!

zashraf1985 (approved):
LGTM! Great Work! This looks good and i am approving it but dont merge it yet. I will do some testing myself to be sure.

PR #149, opened 2022-02-11
fix: React SDK null client handling

opti-jnguyen:
# Summary

Handle React SDK fail catching if ._client returns null when creating new ReactClientSDK instance from `createInstance` function.

# Test Plan

Manually tested thoroughly.
All existing unit tests pass.

# Issues
#134

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -41,6 +41,7 @@ export type OnReadyResult = {
 const REACT_SDK_CLIENT_ENGINE = 'react-sdk';
 const REACT_SDK_CLIENT_VERSION = '2.8.0';
 
+// TODO: Test all functions in ReactSDKClient
 export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserContext'> {
   user: UserInfo;
 
@@ -198,7 +199,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
 
   private isUsingSdkKey: boolean = false;
 
-  private readonly _client: optimizely.Client;
+  private readonly _client: optimizely.Client | null;
 
   // promise keeping track of async requests for initializing client instance
   private dataReadyPromise: Promise<OnReadyResult>;
@@ -210,7 +211,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
   constructor(config: optimizely.Config) {
     this.initialConfig = config;
 
-    this.userPromiseResolver = () => {};
+    this.userPromiseResolver = () => { };
 
     const configWithClientInfo = {
       ...config,
@@ -218,6 +219,11 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       clientVersion: REACT_SDK_CLIENT_VERSION,
     };
     this._client = optimizely.createInstance(configWithClientInfo);
+
+    let rng = Math.random() + 1
+    if (rng > 0) {
+      this._client = null
+    }
 
     this.isClientReady = !!configWithClientInfo.datafile;
     this.isUsingSdkKey = !!configWithClientInfo.sdkKey;
@@ -229,18 +235,25 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       return { success: true };
     });
 
-    this._client.onReady().then(() => {
-      this.isClientReady = true;
-    });
+    if (this._client) {
+      this._client.onReady().then(() => {
+        this.isClientReady = true;
+      });
+    }
 
-    this.dataReadyPromise = Promise.all([this.userPromise, this._client.onReady()]).then(() => {
-      // Client and user can become ready synchronously and/or asynchronously. This flag specifically indicates that they became ready asynchronously.
-      this.isReadyPromiseFulfilled = true;
-      return {
-        success: true,
-        reason: 'datafile and user resolved',
-      };
-    });
+    if (this._client) {
+      this.dataReadyPromise = Promise.all([this.userPromise, this._client.onReady()]).then(() => {
+        // Client and user can become ready synchronously and/or asynchronously. This flag specifically indicates that they became ready asynchronously.
+        this.isReadyPromiseFulfilled = true;
+        return {
+          success: true,
+          reason: 'datafile and user resolved',
+        };
+      });
+    } else {
+      this.dataReadyPromise = Promise.reject(new Error('Optimizely client is null'))
+      this.isReadyPromiseFulfilled = false;
+    }
   }
 
   getIsReadyPromiseFulfilled(): boolean {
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -32,9 +32,12 @@ type OnUserUpdateHandler = (userInfo: UserInfo) => void;
 
 type OnForcedVariationsUpdateHandler = () => void;
 
+type NotReadyReason = 'TIMEOUT' | 'NO_CLIENT';
+
 export type OnReadyResult = {
   success: boolean;
-  reason?: string;
+  reason?: NotReadyReason;
+  reasonDetail?: string;
```
zashraf1985, line 40: A Non binding suggestion: Does it make sense to call it `message` instead of `reasonDetail`? This will make the success case look more inuitive where we dont provide a reason but only reasonDetail right now. But its just a personal preference.. Totally non binding.

src/client.ts
```diff
@@ -498,19 +562,25 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
     const user = this.getUserContextWithOverrides(overrideUserId, overrideAttributes);
 
     if (user.id === null) {
-      logger.info('track for event "%s" not being sent because userId is not set', eventKey);
+      logger.warn('Unable to send tracking event "%s" because User ID is not set', eventKey);
       return;
     }
 
-    return this._client.track(eventKey, user.id, user.attributes, eventTags);
+    this._client.track(eventKey, user.id, user.attributes, eventTags);
```
zashraf1985, line 583: Good catch! This was an unnecessary `return` :+1:

src/client.ts
```diff
@@ -581,7 +651,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
    */
   public removeAllForcedDecisions(): boolean {
     if (!this.userContext) {
-      logger.info("Can't remove a forced decision because the user context has not been set yet");
+      logger.warn('Unable to remove a forced decision because the user context has not been set yet.');
```
zashraf1985, line 654:
```suggestion
      logger.warn('Unable to remove forced decisions because the user context has not been set yet.');
```

src/client.ts
```diff
@@ -608,11 +678,20 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        'Unable to determine if feature "%s" is enabled because Optimizely client failed to initialize.',
+        feature
+      );
+      return false;
+    }
+
     const user = this.getUserContextWithOverrides(overrideUserId, overrideAttributes);
     if (user.id === null) {
-      logger.info('isFeatureEnabled returning false for feature "%s" because userId is not set', feature);
+      logger.warn('Unable to determine if feature "%s" is enabled because User ID is not set', feature);
```
zashraf1985, line 691: Log message nicely rewritten :+1:

src/client.ts
```diff
@@ -635,25 +714,46 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+      logger.warn(
+        'Unable to retrieve feature variables for feature "%s" because Optimizely client failed to initialize.',
+        featureKey
+      );
       return {};
     }
+
     const feature = optlyConfig.featuresMap[featureKey];
     if (!feature) {
+      logger.warn(
```
zashraf1985, line 747: I would suggest to convert that to `info`. Its probably not an unusual behavior.

src/client.ts
```diff
@@ -635,25 +714,46 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        'Unable to retrieve feature variables for feature "%s" because config features map is not set',
+        featureKey
+      );
       return {};
     }
+
     Object.keys(feature.variablesMap).forEach(key => {
       const variable = feature.variablesMap[key];
-      variableObj[variable.key] = this._client.getFeatureVariable(featureKey, variable.key, userId, userAttributes);
+      variableObj[variable.key] = this._client!.getFeatureVariable(featureKey, variable.key, userId, userAttributes);
```
zashraf1985, line 756: The `!` is a good learning for me. Thanks :smile:

src/client.ts
```diff
@@ -675,10 +775,20 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+      logger.warn(
+        'Unable to get feature variable string from feature "%s" because Optimizely client failed to initialize.',
+        feature
+      );
+      return null;
+    }
+
     const user = this.getUserContextWithOverrides(overrideUserId, overrideAttributes);
     if (user.id === null) {
+      logger.warn('Unable to get feature variable string from feature "%s" because User ID is not set', feature);
```
zashraf1985, line 788: I dont know when would this case happen but i will suggest to convert it to info. We might not want existing users to start complaining about new warnings on old use cases.

src/client.ts
```diff
@@ -698,10 +808,21 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        'Unable to get feature variable boolean from feature "%s" because Optimizely client failed to initialize.',
+        feature
+      );
+      return null;
+    }
+
     const user = this.getUserContextWithOverrides(overrideUserId, overrideAttributes);
+
     if (user.id === null) {
+      logger.warn('Unable to get feature variable boolean from feature "%s" because User ID is not set', feature);
```
zashraf1985, line 822: Convert to `info`

src/client.ts
```diff
@@ -721,10 +842,21 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        'Unable to get feature variable integer from feature "%s" because Optimizely client failed to initialize.',
+        feature
+      );
+      return null;
+    }
+
     const user = this.getUserContextWithOverrides(overrideUserId, overrideAttributes);
+
     if (user.id === null) {
+      logger.warn('Unable to get feature variable integer from feature "%s" because User ID is not set', feature);
```
zashraf1985, line 856: make it info

src/client.ts
```diff
@@ -887,24 +1096,66 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
    *  @returns {optimizely.OptimizelyConfig | null} optimizely config
    */
   public getOptimizelyConfig(): optimizely.OptimizelyConfig | null {
+    if (!this._client) {
+      logger.warn('Unable to get Optimizely configuration because Optimizely client failed to initialize.');
```
zashraf1985, line 1100: Lets call it `config`. We never use the full term `configuration` throughout our ecosystem

src/hooks.ts
```diff
@@ -135,18 +135,39 @@ function subscribeToInitialization(
...
-        onInitStateChange({
-          clientReady: true,
-          didTimeout: true,
-        });
-      });
+
+      switch (res.reason) {
+        // Optimizely client failed to initialize.
+        case 'NO_CLIENT':
+          hooksLogger.info(`Client not ready, reason="${res.reasonDetail}"`);
```
zashraf1985, line 142: We can probably convert this one to `warn`. what do you think?

src/hooks.ts
```diff
@@ -431,5 +452,6 @@ export const useDecision: UseDecision = (flagKey, options = {}, overrides = {})
     return (): void => {};
   }, [optimizely.getIsReadyPromiseFulfilled(), options.autoUpdate, optimizely, flagKey, getCurrentDecision]);
 
+  // CHECK: Return clientReady as false & didTimeout as true if ._client is null.
```
zashraf1985, line 455: This was probably your own note that you forgot to delete?

zashraf1985 (changes requested):
Great Work! Its looking really good. I have requested a few minor changes. After that, this one is good to go!

Develliot:
would it also be to have the Provider silently fail if the optimizely prop (ReactSDKClient) is null incase createInstance fails or isn't possible at time of render? 

It's got a update handler to handle optimizely prop being added later

zashraf1985 (approved):
LGTM! Great Work.

src/client.ts
```diff
@@ -231,18 +234,30 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        this.isReadyPromiseFulfilled = true;
+        return {
+          success: true,
+          message: 'Successfully resolved datafile and user information.',
+        };
+      });
+    } else {
+      logger.warn('Unable to resolve datafile and user information because Optimizely client failed to initialize.');
+
+      this.dataReadyPromise = new Promise((resolve, reject) => {
```
fabb, line 253: You could use `Promise.resolve(value)`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

opti-jnguyen: Thanks @fabb - we'll implement that change in a future push!

PR #151, opened 2022-02-19
chore: Adding E2E tests

zashraf1985:
## Summary
React SDK contains unit tests but its not possible to cover many scenarios using those. We have started working on an E2E test suite in a private repo [here](https://github.com/optimizely/react-sdk-e2e-tests). This PR enables this repo to trigger E2E tests on PRs. E2E test suite is not complete yet, so we are not making the check mandatory. Once we have something significant in that repo, we will make this check mandatory.

Since we were working in the travis config file, we cleaned it up a bit and removed the unnecessary markdown lint which has already been removed from other repos.

## Test Plan
Tested the triggering mechanism thoroughly and it appears to be working fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
       before_script:
         - mkdir $HOME/travisci-tools && pushd $HOME/travisci-tools && git init && git pull https://$CI_USER_TOKEN@github.com/optimizely/travisci-tools.git && popd
       script:
-        - $HOME/travisci-tools/trigger-script-with-status-update.sh
+        - $HOME/travisci-tools/trigger-script-with-status-update.sh main
       after_success: travis_terminate 0
 
     - stage: 'Publish'
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: generic
 os: linux
 
 stages:
-  - 'Lint markdown files'
-  - 'Test'
   - 'Integration tests'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,6 +17,21 @@ jobs:
       addons:
         srcclr: true
 
+    - stage: 'Integration tests'
+      cache: false
```
msohailhussain, line 21: can you please add why you added this one?

.travis.yml
```diff
@@ -25,6 +17,21 @@ jobs:
...
+      language: minimal
+      env:
+        SDK=react
+        SDK_BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
+        REPO_SLUG=optimizely/react-sdk-e2e-tests
```
msohailhussain, line 26: Please specify, it's other than conventional naming that's why specifying REPO_SLUG otherwise, we don't need it.

msohailhussain (approved):
lgtm

.travis.yml
zashraf1985: Copy pasted from javascript SDK ([here](https://github.com/optimizely/javascript-sdk/blob/8eedd63296fc722e737dd7f7a0ac77e43018a8c8/.travis.yml#L50)), i dont know travis CI much.. didnt want to take any risks making modifications on my own :)

Issue #152, opened 2022-03-01
Hooks to include error in return instead of throw errors


Develliot:
Currently if client isn't ready hooks will throw an error. 

I think the API would be much nicer if it simply returned an error like useSWR does, then I wouldn't need to wrap everything in a try catch to stop my app falling over.

return {decision: OptimizelyDecision | null, clientReady: boolean, didTimeout: boolean, isLoading: boolean, error: Error | null}

PR #154, opened 2022-03-09
fix: Hooks gracefully return with an error message when optimizely prop is null

opti-jnguyen:
## Summary

Enables usage of hooks within the Optimizely Provider without passing in the `optimizely` prop.

Previously, attempting to use hooks within an Optimizely Provider with a missing `optimizely` prop led to an error being thrown resulting in a crash.

With this change, hooks now return pessimistic results when the `optimizely` prop is `null`.

Addresses this issue: 
- https://github.com/optimizely/react-sdk/issues/152

Also related:
- https://github.com/optimizely/react-sdk/issues/134

Note:
- Instead of returning an "Error" message like requested, we will defer to SDK users to handle the pessimistic values. 
- In certain edge cases, this may be a potentially breaking change for users that expect hooks to throw an error when the `optimizely` prop is null.

---

## Test Plan

Refactored tests that expect hooks to throw errors when Optimizely provider is not present.

### DIFF
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018-2019, Optimizely
+ * Copyright 2022, Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -66,7 +66,15 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
     }
 
     if (finalUser) {
-      optimizely.setUser(finalUser);
+      if (!optimizely) {
+        logger.warn(`Unable to set user ${finalUser} because optimizely object does not exist.`)
+      } else {
+        try {
+          optimizely.setUser(finalUser);
+        } catch (err) {
+          logger.warn(`Unable to set user ${finalUser} because passed in optimizely object does not contain the setUser function.`)
+        }
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/Provider.tsx
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018-2019, Optimizely
+ * Copyright 2022, Optimizely
```
opti-jnguyen, line 2: 2018-2019, 2022

src/hooks.ts
```diff
@@ -365,11 +375,22 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
   const { optimizely, isServerSide, timeout } = useContext(OptimizelyContext);
+
+  const overrideAttrs = useCompareAttrsMemoize(overrides.overrideAttributes);
+
   if (!optimizely) {
-    throw new Error('optimizely prop must be supplied via a parent <OptimizelyProvider>');
+    hooksLogger.warn(`Unable to use decision ${flagKey}. optimizely prop must be supplied via a parent <OptimizelyProvider>`);
+    return [
+      createFailedDecision(flagKey, 'Optimizely SDK not configured properly yet.', {
+        id: overrides.overrideUserId || null,
```
opti-jnguyen, line 385: Return null for user items bc user depends on Optimizely object

zashraf1985 (approved):
Looks Great! Can you change the PR description format to what we usually use like "Summary", "Test Plan" etc.

PR #159, opened 2022-04-27
Fix: Upgrade to react 18

shaharyar123:
## Summary
1. Updates React SDK to use React 18 and fixed related typescript issues.
2. Replaces `enzyme` with `react testing library` to make unit tests work with React 18.

## Test Plan
- Updated many unit tests to use React Testing Library instead of Enzyme.
- E2E Tests also Pass.


.github/workflows/integration_test.yml
```diff
@@ -45,5 +45,5 @@ jobs:
         TRAVIS_COM_TOKEN: ${{ secrets.TRAVIS_COM_TOKEN }}
         EVENT_MESSAGE: ${{ github.event.message }}
         HOME: 'home/runner'
-      run: |
```
zashraf1985, line 48: This looks like an unnecessary change. can you re add the space here so that it does not appear in the PR diff

src/hooks.spec.tsx
```diff
@@ -163,51 +167,48 @@ describe('hooks', () => {
...
       readySuccess = false;
 
-      const component = Enzyme.mount(
+      render(
         <OptimizelyProvider optimizely={optimizelyMock}>
           <MyExperimentComponent options={{ timeout: mockDelay }} />
         </OptimizelyProvider>
       );
-      expect(component.text()).toBe('null|false|false'); // initial render
+      await waitFor(() => expect(screen.getByTestId('result')).toHaveTextContent('null|false|false'));
```
zashraf1985, line 201: Can you add the `// initial render` comment here again

zashraf1985 (changes requested):
Looks Great! Just two minor change requests

zashraf1985:
@opti-jnguyen Can you take a look when you get some time.

zashraf1985 (approved):
LGTM! Great work @shaharyar123

opti-jnguyen (approved):
Seems fine when testing locally - LGTM!

PR #182, opened 2022-08-23
Add ability to conditionally render experimentation hooks

PepijnSenders:
## Description

We'd like to be able to render `useExperiment()` and the other hooks conditionally. This PR will add the ability for people to run `useExperiment('experiment_key', { disabled: true })` to skip the execution of the hook.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -34,6 +34,7 @@ enum HookType {
 type HookOptions = {
   autoUpdate?: boolean;
   timeout?: number;
+  disabled?: boolean;
 };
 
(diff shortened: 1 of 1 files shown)


mikechu-optimizely:
This is a solid addition.  I've added a Jira ticket for this (FSSDK-9627; need to add tests, etc)

I see you've already signed the Contributor License Agreement.  Thanks.

junaed-optimizely:
This indeed could be a great addition to the hooks. However hook structure has been changed, and this version of the code goes agains the [Rules of Hook](https://react.dev/reference/rules/rules-of-hooks)

I would suggest work on the latest `master` and follow the **Rules of Hook** accordingly. 

Thank you!

PR #184, opened 2022-10-05
added useCallback to getCurrentDecision in useDecision hook

monahans:
adding useCallback here fixed a few things:
1. Bug: if you used multiple useDecision hooks with { autoUpdate: true }, the second one would not autoUpdate (you had to fully reload the app to get the decision to update)
2. Before it was redefining getCurrentDecision on every rerender of the useDecision hook, which decreased performance


mikechu-optimizely:
This seems to relate to #196 and (now) internal ticket FSSDK-9624

junaed-optimizely:
@monahans, this has been added in #273 with so many other improvements. Thanks for the suggestion. Closing this!

PR #204, opened 2023-07-04
[FSSDK-9471] [React] Fetch segments in onready

rafinutshaw-optimizely:
## Summary
- Add fetchqualifiedsegments options params to onReady client method
- Fetch qualified segments in onready if usercontext is available
- 
## Test plan
- fetchqualifiedsegment method is called in onReady

## Issues
- https://jira.sso.episerver.net/browse/FSSDK-9471

### DIFF
--- a/src/client.spec.ts
+++ b/src/client.spec.ts
@@ -172,6 +172,15 @@ describe('ReactSDKClient', () => {
         });
       });
 
+      it('does not call fetchqualifiedsegments on dataReadyPromise success = false', async () => {
+        instance.setUser({
+          id: 'user12345',
+        });
+        jest.spyOn(instance, 'fetchQualifiedSegments').mockImplementation(async () => true);
+        await instance.onReady();
+        expect(instance.fetchQualifiedSegments).toHaveBeenCalledTimes(0);
+      });
+
       it('fulfills the returned promise with success: false when a user is set', async () => {
         instance.setUser({
           id: 'user12345',
--- a/src/client.spec.ts
+++ b/src/client.spec.ts
@@ -142,6 +142,15 @@ describe('ReactSDKClient', () => {
       expect(result.success).toBe(false);
     });
 
+    it('calls fetchqualifiedsegments when a user is set', async () => {
+      instance.setUser({
+        id: 'user12345',
+      });
+      jest.spyOn(instance, 'fetchQualifiedSegments').mockImplementation(async () => true);
+      await instance.onReady();
+      expect(instance.fetchQualifiedSegments).toHaveBeenCalledTimes(1);
+    });
+
     it('fulfills the returned promise with success: true when a user is set', async () => {
       instance.setUser({
         id: 'user12345',
--- a/src/client.ts
+++ b/src/client.ts
@@ -302,7 +302,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     });
 
     return Promise.race([this.dataReadyPromise, timeoutPromise]).then(async res => {
-      await this.fetchQualifiedSegments(config?.segmentOptions);
+      if (res.success) await this.fetchQualifiedSegments(config?.segmentOptions);
       clearTimeout(timeoutId);
       return res;
     });
--- a/src/client.ts
+++ b/src/client.ts
@@ -44,7 +44,7 @@ const REACT_SDK_CLIENT_VERSION = '2.9.2';
 export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserContext'> {
   user: UserInfo;
 
-  onReady(opts?: { timeout?: number }): Promise<any>;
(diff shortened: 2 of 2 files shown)


src/client.spec.ts
```diff
@@ -142,6 +142,15 @@ describe('ReactSDKClient', () => {
       expect(result.success).toBe(false);
     });
 
+    it('calls fetchqualifiedsegments when a user is set', async () => {
```
mikechu-optimizely, line 145: Should we add some additional tests for the negative cases like 
- when we don't have a userContext or 
- when there's a timeout and we verify that there was a `message` for 'failed to initialize onReady before timeout, either the datafile, user info or qualified segments was not set before the timeout'

src/client.ts
```diff
@@ -281,26 +281,28 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
-            'failed to initialize onReady before timeout, either the datafile or user info was not set before the timeout',
+            'failed to initialize onReady before timeout, either the datafile, user info or qualified segments was not set before the timeout',
           dataReadyPromise: this.dataReadyPromise,
         });
       }, timeout) as any;
     });
 
-    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(res => {
+    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(async res => {
+      await this.fetchQualifiedSegments(config?.segmentOptions);
```
mikechu-optimizely, line 305: Do we need testing around a failure for `fetchQualifiedSegments()`?

mikechu-optimizely: ...or maybe we already cover it.

mikechu-optimizely (approved):
LGTM, with no showstoppers from my perspective.  I have a couple of questions below.

src/client.ts
jaeopt, line 305: `fetchQualifiedSegment` should be called only on datafile success not on timeout. Also can you check if this the caller is blocked until `fetchQualifiedSegments` completed. It looks like async call.

jaeopt (changes requested):
Questions about fetch timing.

src/client.spec.ts
rafinutshaw-optimizely: - we already have a userContext check inside fetchqualified method
- Updated changes for this

src/client.ts
rafinutshaw-optimizely: yes failure for fetchQualifiedSegments has been cover on fetchQualifiedSegments method itself. So we can just fire and forget

rafinutshaw-optimizely: Thanks. added a check timeout check on this. And yes this the caller is blocked until fetchQualifiedSegments completed

src/client.ts
```diff
@@ -294,13 +296,14 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
-            'failed to initialize onReady before timeout, either the datafile or user info was not set before the timeout',
+            'failed to initialize onReady before timeout, either the datafile, user info or qualified segments was not set before the timeout',
           dataReadyPromise: this.dataReadyPromise,
         });
       }, timeout) as any;
     });
 
-    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(res => {
+    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(async res => {
+      if (res.success) await this.fetchQualifiedSegments(config?.segmentOptions);
```
mikechu-optimizely, line 306: Can we use two lines for this and enclose in { } ?

mikechu-optimizely (approved):
LGTM

src/client.ts
```diff
@@ -294,13 +296,16 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
           success: false,
           reason: 'TIMEOUT',
           message:
-            'failed to initialize onReady before timeout, either the datafile or user info was not set before the timeout',
+            'failed to initialize onReady before timeout, either the datafile, user info or qualified segments was not set before the timeout',
```
jaeopt, line 299: I believe this timeout is covering datafile ready timeout, not covering "qualifed segments" failure. The original message makes more sense.

src/client.ts
```diff
@@ -294,13 +296,16 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
           dataReadyPromise: this.dataReadyPromise,
         });
       }, timeout) as any;
     });
 
-    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(res => {
+    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(async res => {
+      if (res.success) {
+        await this.fetchQualifiedSegments(config?.segmentOptions);
```
jaeopt, line 307: What if fetch fails with network error, timeout, etc? res should be updated accordingly?

jaeopt (changes requested):
The flag looks good. A couple of more questions.

rafinutshaw-optimizely:
Closing this due to change in requirement.

PR #205, opened 2023-07-20
[FSSDK-9492] [React] Fetch segments in setuser

rafinutshaw-optimizely:
## Summary
- Fetch qualifiedsegments in setuser internally, so that it gets called everytime user is updated

## Test plan
- add test for fetchqualifiedsegment method being called internally for each setuser call 

## Issues
- https://jira.sso.episerver.net/browse/FSSDK-9492

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -41,6 +41,11 @@ export type OnReadyResult = {
 const REACT_SDK_CLIENT_ENGINE = 'react-sdk';
 const REACT_SDK_CLIENT_VERSION = '2.9.2';
 
+const default_user: UserInfo = {
+  id: null,
+  attributes: {},
+};
+
 export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserContext'> {
   user: UserInfo;
 
@@ -203,10 +208,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
   private dataReadyPromise: Promise<OnReadyResult>;
 
   public initialConfig: optimizely.Config;
-  public user: UserInfo = {
-    id: null,
-    attributes: {},
-  };
+  public user: UserInfo = default_user;
 
   /**
    * Creates an instance of OptimizelyReactSDKClient.
@@ -350,6 +352,9 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
   public async setUser(userInfo: UserInfo): Promise<void> {
     this.isUserPromiseResolved = false;
     this.isUserReady = true;
+
+    //reset user info
+    this.user = default_user;
 
     this.user.id = userInfo.id;
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -347,7 +347,10 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     return await this.userContext.fetchQualifiedSegments(options);
   }
 
-  public setUser(userInfo: UserInfo): void {
+  public async setUser(userInfo: UserInfo): Promise<void> {
+    this.isUserReady = false;
+    this.isUserPromiseResolved = false;
+
     // TODO add check for valid user
     if (userInfo.id) {
       this.user.id = userInfo.id;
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -347,26 +349,30 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+        'Unable to create user context for user id "%s" because Optimizely client failed to initialize.',
+        this.user.id
+      );
     }
 
     if (userInfo.attributes) {
       this.user.attributes = userInfo.attributes;
     }
 
+    await this.fetchQualifiedSegments();
```
jaeopt, line 374: Error handling? 
Log error message if this returns false? Should we continue on error?

jaeopt: We also need support for options for fetchQualifedSegments.

src/client.ts
```diff
@@ -347,26 +349,30 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
-          this.user.id
-        );
-      }
+    //reset user info
+    this.user = default_user;
+
+    this.user.id = userInfo.id;
+
+    if (this._client) {
+      this.userContext = this._client.createUserContext(userInfo.id ?? undefined, userInfo.attributes);
```
jaeopt, line 362: I do not see new test cases covering null userId cases.

jaeopt (changes requested):
Looks good. A couple of changes suggested.

src/client.ts
rafinutshaw-optimizely: the this._client.createUserContext only accepts string or undefined. and this undefined is then replaced with viud in javascript

rafinutshaw-optimizely: hence whatever falsy value the client may provide the react-sdk will be passing them as undefined

src/client.ts
rafinutshaw-optimizely: the error handling for fetchqualifiedsegment is done independently in the fetchqualifiedsegment method itself. 

for the options support, since setuser may be a new user everytime, options dont have any purpose here. If its a new user a new calll is made instead of reading from the cache, the userid being the deciding key here.

jaeopt: I understand all errors logged in js core. Wondering we should continue to resolve the promise when we get false for `fetchQualifedSegments`?

src/client.ts
```diff
@@ -230,17 +230,28 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
       this._client.onReady().then(() => {
         this.isClientReady = true;
       });
 
-      this.dataReadyPromise = Promise.all([this.userPromise, this._client!.onReady()]).then(() => {
+      this.dataReadyPromise = Promise.all([this.userPromise, this._client!.onReady()]).then(res => {
+        if (!res[0].success) {
+          return {
+            success: false,
+            reason: 'QUALIFIED_SEGMENTS_NOT_READY',
```
jaeopt, line 250: What about more generic "USER_NOT_READY" here for future extension? 
`message` below will have details "segments not ready".

src/client.ts
```diff
@@ -230,17 +230,28 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
       this.userPromiseResolver = resolve;
-    }).then(() => {
-      this.isUserReady = true;
-      return { success: true };
+    }).then(isUserResolved => {
+      if (isUserResolved) {
+        this.isUserReady = true;
+        return { success: true };
+      }
+      return { success: false, message: 'Failed to fetch qualified segments.' };
```
jaeopt, line 238: Same here. Can we get the reason of user resolve failure from the caller?  
There can be other reasons in future for failure.

jaeopt (changes requested):
Looks good! A couple of suggestions for message improvement.

src/client.ts
```diff
@@ -230,12 +233,9 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
-    }).then(isUserResolved => {
-      if (isUserResolved) {
-        this.isUserReady = true;
-        return { success: true };
-      }
-      return { success: false, message: 'Failed to fetch qualified segments.' };
+    }).then((result: ResolveResult) => {
+      this.isUserReady = result.success;
+      return result;
     });
```
jaeopt: I see resolver is set and called later. Wondering the value of this promise setting instead of calling line 237-238?

jaeopt (changes requested):
One more clarification. Other than that, all looks good!

jaeopt (approved):
I see why we need it! LGTM

PR #206, opened 2023-08-02
[FSSDK-9563] [React] Fix onready segments

rafinutshaw-optimizely:
## Summary
- OnReady should only fetchqualifiedsegments if user is ready
- Setuser should only fetchqualified segments if js-client is ready

## Test plan
- Current test cases should pass

## Issues
- https://jira.sso.episerver.net/browse/FSSDK-9563

### DIFF
--- a/src/client.spec.ts
+++ b/src/client.spec.ts
@@ -17,10 +17,10 @@ jest.mock('@optimizely/optimizely-sdk');
 jest.mock('./logger', () => {
   return {
     logger: {
-      warn: jest.fn(() => () => {}),
-      info: jest.fn(() => () => {}),
-      error: jest.fn(() => () => {}),
-      debug: jest.fn(() => () => {}),
+      warn: jest.fn(() => () => { }),
+      info: jest.fn(() => () => { }),
+      error: jest.fn(() => () => { }),
+      debug: jest.fn(() => () => { }),
     },
   };
 });
@@ -251,7 +251,7 @@ describe('ReactSDKClient', () => {
       expect(onUserUpdateListener).toBeCalledTimes(1);
     });
 
-    it('calls fetchqualifiedsegements internally on each setuser call', async () => {
+    it('does not call fetchqualifiedsegements on setUser if onready is not calleed initially', async () => {
       const instance = createInstance(config);
       jest.spyOn(instance, 'fetchQualifiedSegments').mockImplementation(async () => true);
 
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -309,8 +306,18 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       }, timeout) as any;
     });
 
-    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(res => {
+    return Promise.race([this.dataReadyPromise, timeoutPromise]).then(async res => {
       clearTimeout(timeoutId);
+      if (res.success) {
+        const isSegmentsFetched = await this.fetchQualifiedSegments();
```
jaeopt, line 312: Can you confirm if JS SDK returns true when ODP is not integrated or disabled. It should continue to work as success for backward compatibility.

jaeopt: I see `fetchQualifiedSegments` called both from setUser and here. Wondering if they are not called twice when SDK starts. Can we add a comment about the flow?

jaeopt (changes requested):
A good catch! A couple of clarifications.

src/client.ts
rafinutshaw-optimizely: it does return true when ODP is not integrated or disabled.

rafinutshaw-optimizely: for the flow, to initialize the sdk we need to call the js onready which is awaited on the onready method. so for the initial call we need to call segment fetch on onready and the on the setuser segment call will execute on any following setuser calls after initialization. 
and there is a check on setuser line:387 so that during initializing segment fetch is not called twice.

src/client.spec.ts
```diff
@@ -251,19 +251,35 @@ describe('ReactSDKClient', () => {
...
+      await instance.setUser({
+        id: 'xxfueaojfe8&87',
+      });
+
       await instance.setUser({
         id: 'xxfueaojfe8&87',
       });
 
-      expect(instance.fetchQualifiedSegments).toBeCalledTimes(2);
+      expect(instance.fetchQualifiedSegments).toBeCalledTimes(3);
```
mikechu-optimizely, line 282: Should this only be 2 since `setUser` is called 2x after the `onReady()`? ...or maybe I'm confused about the wording of the test name. 😕

jaeopt (approved):
LGTM

src/client.spec.ts
mikechu-optimizely: Got it.  This is where a `// comment` may be a useful addition for the future.  

Thanks for explaining.

PR #221, opened 2023-10-27
[FSSDK-9726] AAT gap fill

mikechu-optimizely:
## Summary

- Add missing AAT/ODP functions exposed in the JS SDK but not available to React
- Quality of life improvements

## Test plan

- New unit tests added for coverage
- Existing unit and e2e tests are expected to pass.

## Issues

- [FSSDK-9726](https://jira.sso.episerver.net/browse/FSSDK-9726)


src/client.ts
```diff
@@ -180,9 +180,13 @@ export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserConte
   getForcedDecision(decisionContext: optimizely.OptimizelyDecisionContext): optimizely.OptimizelyForcedDecision | null;
 
   fetchQualifiedSegments(options?: optimizely.OptimizelySegmentOption[]): Promise<boolean>;
+
+  getQualifedSegments(): string[] | null;
```
jaeopt, line 184: This must be an API for UserContext. Do we need this at client level?

src/client.ts
```diff
@@ -343,8 +347,8 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       }
 
       if (userInfo.id) {
-        userContext = this._client.createUserContext(userInfo.id, userInfo.attributes);
-        return userContext;
+        this.userContext = this._client.createUserContext(userInfo.id, userInfo.attributes);
```
jaeopt, line 350: Do we want to replace existing userContext if non-matching userId is requested?
It can be a breaking change.

src/client.ts
```diff
@@ -367,6 +371,15 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     return await this.userContext.fetchQualifiedSegments(options);
   }
 
+  public getQualifedSegments(): string[] | null {
```
jaeopt, line 374: Do we need this extra API? They can get/set qualifiedSegments after getUserContext().

jaeopt (changes requested):
Looks good. A couple of clarifications.

src/client.ts
mikechu-optimizely: During bug bash React Native, I had no way to verify what qualified segments were given.  We're hiding this from React but not to JS. My thought was that we should have a way to view. 

This was part of the gap. 

Perhaps a call is good to review tomorrow?

mikechu-optimizely: After sleeping on it, I think I understand. 

Are you looking for something like:
`const qualifedSetments = optimizelyClient.getUserContextInstance({id: 'matjaz-user-2'})?.qualifiedSegments;`

src/client.ts
raju-opti: yeah, I have the same question as Jae

src/client.spec.ts
```diff
@@ -1624,4 +1624,82 @@ describe('ReactSDKClient', () => {
...
+    });
+
+    it('should return a valid vuid', async () => {
+      const validVuid = 'vuid_8de3bb278fce47f6b000cadc1ac';
+      const mockGetVuid = mockInnerClient.getVuid as jest.Mock;
+      mockGetVuid.mockReturnValue(validVuid);
+
+      const vuid = instance.getVuid();
+
+      expect(vuid).toMatch(vuidFormat);
```
raju-opti, line 1670: should we also test that the mocked vuid is returned and remove the previous two tests?

raju-opti (changes requested):

src/client.ts
mikechu-optimizely: I've updated

src/client.ts
mikechu-optimizely: Correction made.

mikechu-optimizely (commented):
Developer re-reviewed

src/client.ts
```diff
@@ -181,7 +181,7 @@ export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserConte
 
   fetchQualifiedSegments(options?: optimizely.OptimizelySegmentOption[]): Promise<boolean>;
 
-  getQualifedSegments(): string[] | null;
+  getUserContextInstance(userInfo: UserInfo): optimizely.OptimizelyUserContext | null;
```
jaeopt, line 184: Not sure but wondering if we can use "getUser" as a pair with existing "setUser" method.

jaeopt (changes requested):
A clarifying question. Other than that, all looks good!

src/client.spec.ts
raju-opti: I mean adding this assertion:

`expect(vuid).toEqual(validVuid);`

and removing previous two test that the js sdk getVuid() is called

mikechu-optimizely: Thanks for clarifying

src/client.ts
mikechu-optimizely: Oh. That makes sense. I noticed that the existing `public getUserContextInstance()` wasn't on the interface and only used internally on the concrete implementation, so that's why I chose it.

Do you want me to remove it from the interface and add a `getUser`?

jaeopt: Yes, simple getter `getUser() { return this.currentContext }` instead of `getUserContextInstance()` makes more sense to me.

mikechu-optimizely: Hi @jaeopt,

I've pushed the changes per our call.  Thanks for another review.

raju-opti (approved):
LGTM

mikechu-optimizely:
@jaeopt Can I get your approval when you have a moment?

jaeopt (approved):
LGTM

Issue #225, opened 2023-11-30
[BUG] Default <OptimizelyVariation> always rendered when not final child of <OptimizelyExperiment>


alexparish:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

2.9.2

### Current Behavior

The `<OptimizelyVariation>` component with default attribute is rendered in all cases when it appears as the first child component of `<OptimizelyExperiment>`:

```
<OptimizelyExperiment experiment="simple_experiment">
  <OptimizelyVariation default>
    default
  </OptimizelyVariation>
  <OptimizelyVariation variation="control">
    control
  </OptimizelyVariation>
  <OptimizelyVariation variation="variation">
    variation
   </OptimizelyVariation>
</OptimizelyExperiment>
```

### Expected Behavior

The correct chosen variation should be rendered. Which variation is rendered which should be agnostic of the position of the default `<OptimizelyVariation>` component within the `<OptimizelyExperiment>` component.

### Steps To Reproduce

Configure an experiment with the following code:

```
<OptimizelyExperiment experiment="simple_experiment">
  <OptimizelyVariation default>
    default
  </OptimizelyVariation>
  <OptimizelyVariation variation="control">
    control
  </OptimizelyVariation>
  <OptimizelyVariation variation="variation">
    variation
   </OptimizelyVariation>
</OptimizelyExperiment>
```

Ensure the visitor is eligible for the experiment and that the control or variation are configured to be randomly chosen as the variation. Enable the experiment.

The `<OptimizelyVariation>` component with default attribute will always be rendered even when the experiment is enabled and the experiment decision is "control" or "variation".

### React Framework

18.2.0

### Browsers impacted

All

### Link

https://codesandbox.io/p/sandbox/optimizely-demo-forked-rsjkml?file=%2Fsrc%2Findex.jsx

### Severity

Minor issue

mikechu-optimizely:
Oh shoot. You said that in the title and current behaviour. Sorry Mate.  

Let me read the Issue (better) and SDK source and get back to you.

mikechu-optimizely:
For reference, I've created an internal ticket FSSDK-9839 for this.

alexparish:
That's great thanks. I wonder if it's worth tackling https://github.com/optimizely/react-sdk/issues/86 at the same time given they are in the same area of the codebase?

mikechu-optimizely:
Nice. Agreed.

mikechu-optimizely:
[...]

PR #226, opened 2023-12-05
[FSSDK-9839] fix: rendering `default` OptimizelyVariation when not last

mikechu-optimizely:
## Summary
- Fix for #225 where `OptimizelyVariations` with `default` must always be the last child component within an `OptimizelyExperiment`

## Test plan

- Added new tests for coverage
- All existing unit and e2e tests are still expected to pass

## Issues
- FSSDK-9839


raju-opti (approved):
Looks good. Thanks

PR #229, opened 2023-12-28
[FSSDK-9654] fix: Default to VUID without user(id) as Provider prop

mikechu-optimizely:
## Summary
- When `OptimizelyProvider` is used without the `user?` or `userId?` props, we should retrieve and use the VUID for `optimizely.setUser()`
- Refactored logic from `constructor()` of provider into lifecycle method `componentDidMount()`
- Test updates to support `getVuid()`
- Added types to remove warns.

## Test plan

- Existing and updated unit tests are expected to pass
- Integration and e2e test should continue to pass
- I'd like to review with @jaeopt for AAT function reqs.

## Issues
- FSSDK-9654

### DIFF
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -45,7 +45,11 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
     super(props);
   }
 
-  async componentDidMount() {
+  componentDidMount(): void {
+    this.setUserInOptimizely();
+  }
+
+  async setUserInOptimizely(): Promise<void> {
     const { optimizely, userId, userAttributes, user } = this.props;
 
     if (!optimizely) {
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -57,7 +57,6 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
       return;
     }
 
-    //await optimizely.onReady(); // causes problems with hooks
     let finalUser: UserInfo | null = null;
 
     if (user) {
@@ -84,9 +83,10 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
 
     if (finalUser) {
       try {
+        await optimizely.onReady();
         await optimizely.setUser(finalUser);
       } catch {
-        logger.error('Unable to set user. Prop optimizely object does not contain the setUser function.');
+        logger.error('Error while trying to set user.');
       }
     }
   }
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -37,7 +37,7 @@ interface OptimizelyProviderProps {
 
 interface OptimizelyProviderState {
   userId: string;
-  attributes: { [key: string]: string; } | undefined;
+  attributes: { [key: string]: string } | undefined;
 }
 
 export class OptimizelyProvider extends React.Component<OptimizelyProviderProps, OptimizelyProviderState> {
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -43,10 +43,14 @@ interface OptimizelyProviderState {
 export class OptimizelyProvider extends React.Component<OptimizelyProviderProps, OptimizelyProviderState> {
   constructor(props: OptimizelyProviderProps) {
     super(props);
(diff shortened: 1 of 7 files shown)


mikechu-optimizely:
I'll re-open later when I've more progress.

mikechu-optimizely:
I think I'm actually ready for a review of this @raju-opti and @jaeopt. 

Sorry, this took so long to untangle this new anon paradigm using vuid.

mikechu-optimizely:
Developer review completed.

mikechu-optimizely:
I have an e2e test config set up and able to demo effect.

src/client.ts
```diff
@@ -330,95 +309,62 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
-        'Unable to get user context because user was not set.'
-      );
+      logger.warn('Unable to get user context. User context not set.');
       return null;
     }
 
     return this.userContext;
   }
 
-  public getUserContextInstance(userInfo: UserInfo): optimizely.OptimizelyUserContext | null {
```
jaeopt, line 351: same question about removing public method for backward compatibility?

src/client.ts
```diff
@@ -235,31 +231,18 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     this.isClientReady = !!configWithClientInfo.datafile;
     this.isUsingSdkKey = !!configWithClientInfo.sdkKey;
 
-    this.userPromise = new Promise(resolve => {
```
jaeopt, line 238: I cannot follow all the dependences, but it looks like you simplified a lot. Looks great! 
Just want to make sure this does not break the case for async user setting.

jaeopt (changes requested):
The fix for using vuid for null userId looks good to me. 
A couple of clarifications regarding backward compatibility.

src/client.ts
```diff
@@ -181,7 +181,11 @@ export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserConte
 
   fetchQualifiedSegments(options?: optimizely.OptimizelySegmentOption[]): Promise<boolean>;
 
-  getUserContext(): optimizely.OptimizelyUserContext | null;
```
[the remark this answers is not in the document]
mikechu-optimizely: It's a good callout. I think I wanted to clarify that the consumer is receiving the current user context, especially in the light of separating the responsibilities of `getUserContextInstance`. 

I'm up for renaming it back to `getUserContext` and `setUserContext` for consistency across SDKs.

src/client.ts
mikechu-optimizely: It's also a good callout.  This `getUserContextInstance` was trying to do too much in one call. I've separated out the responsibilities into `getCurrentUserContext` (or just `getUserContext` if you prefer), `setCurrentUserContext`, and `makeUserContextInstance` so that there's clarity on what each does.  

This major version seemed like a good time to update for clarity with less concern about backwards compatibility.  

You are reminding me, though, that I need to make the matching change suggestions in @trishahanlon 's  React docs.

src/client.ts
mikechu-optimizely: You and me both Sir.  It took too long to understand the Promise web. I'd love to unravel this more but I would rather not scope creep past the GA release.

Yes, the goal code allows for latent setting of the user in the user context leaning on the VUID in the absence of a real userId. 

I have a devcontainer ready and happy to demo the effect.

mikechu-optimizely (commented):
Thanks @jaeopt . I've replied to your comments.

@raju-opti I'd love your double sign off too on this PR.

jaeopt (approved):
LGTM

src/Provider.tsx
```diff
@@ -42,9 +43,20 @@ interface OptimizelyProviderState {
 export class OptimizelyProvider extends React.Component<OptimizelyProviderProps, OptimizelyProviderState> {
   constructor(props: OptimizelyProviderProps) {
     super(props);
-    const { optimizely, userId, userAttributes, user } = props;
+  }
+
+  componentDidMount(): void {
```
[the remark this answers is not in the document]
mikechu-optimizely: Internal ticket opened and put into this sprint.

PR #248, opened 2024-03-05
[FSSDK-9969] replace uglify plugin with terser

raju-opti:
## Summary
- rollup-plugin-uglify has not been updated since 2019 and the last version has vulnerable dependencies causing dependabot alerts. Replaced this with terser plugin instead.

## Issues
- FSSDK-9969


mikechu-optimizely (approved):
LGTM. Thanks 🙏

Issue #236, opened 2024-01-29
[BUG] ReactSDKClient setUser() is not working


dtrenz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

`3.0.0`

### Current Behavior

When attempting to set the user ID on the Optimizely ReactSDK client using `setUser()` I see the following error when implementing a feature flag:
```
Unable to determine if feature "<feature key>" is enabled because User ID is not set
```

### Expected Behavior

Setting a user ID using `setUser()` should set the usr ID on the client.

### Steps To Reproduce

```tsx
import { createInstance, OptimizelyProvider } from "@optimizely/react-sdk";
 
const App = () => {
  const optimizelyClient = createInstance({
    sdkKey: "<key>",
  });

  optimizelyClient.setUser({ id: "12345" });

  return (
    <OptimizelyProvider optimizely={optimizelyClient}>
      <Children />
    </OptimizelyProvider>
  );
}
```

...then, in a child component:

```tsx
  const [isEnabled] = useFeature("<feature key>");
```

### React Framework

React 18.2.0 (Create React App + react-app-rewired)

### Browsers impacted

Reproducible in both Chrome and Safari. No other browsers were tested. The issue does not seem to be browser-specific.

### Logs

![Screenshot 2024-01-29 at 4 12 02 PM](https://github.com/optimizely/react-sdk/assets/639833/ab7c8b67-78d8-4232-85ea-333c5c36d62c)

### Workaround/Solution

I'm forced to abandon this approach and have to set the user ID via the `user` prop in `<OptimizelyProvider>`.

### Recent Change

No

mikechu-optimizely:
Thanks for the report @dtrenz . Let me spin up a similar configuration as you've provided and try out.  I'll work on #235 at the same time.

mikechu-optimizely:
I've added an internal ticket FSSDK-9985 into our next sprint.

ricky-harrys:
I'm seeing this issue as well. It's being evaluated before the SDK is ready but it should be available in `<OptimizelyProvider />` at this point.

![Screenshot 2024-02-01 at 11 50 11 AM](https://github.com/optimizely/react-sdk/assets/107486935/1efe3c75-9f87-4ed0-b563-61aa9e574efe)

danieleloscozzese:
[...]

Issue #235, opened 2024-01-29
[BUG] "ReactSDK: Error initializing client. The core client or user promise(s) rejected."


dtrenz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

`3.0.0`

### Current Behavior

When following the instructions for implementing the React SDK for feature flags, I get this error in the JS console: 
```
ReactSDK: Error initializing client. The core client or user promise(s) rejected.
```

### Expected Behavior

No error is encountered when initializing the Optimizely React SDK.

### Steps To Reproduce

```tsx
import { createInstance, OptimizelyProvider } from "@optimizely/react-sdk";
 
const App = () => (
  <OptimizelyProvider
    optimizely={createInstance({  sdkKey: "<key>" })}
    user={{ id: "12345" }}
  >
   <Children />
  </OptimizelyProvider>
);
```

...then, in a child component:

```tsx
  const [isEnabled] = useFeature("<feature key>");
```

### React Framework

React 18.2.0 (Create React App + react-app-rewired)

### Browsers impacted

Reproducible in both Chrome and Safari. No other browsers were tested. The issue does not seem to be browser-specific.

### Logs

![image](https://github.com/optimizely/react-sdk/assets/639833/d337c79b-b6fc-422f-884f-0b4bd1c377bc)

### Severity

Minor issue

### Workaround/Solution

No, but the SDK seems to still work despite this error, so I suppose I could ignore it?

### Recent Change

This is a brand new integrations, so no.

mikechu-optimizely:
I would complete expect No Errors even though the SDK is functional. Lemme hop on this.

mikechu-optimizely:
Created internal ticket FSSDK-9984 and put it into next sprint.

dominikjasek:
I am having same issue: here's a reproduction codesandbox: https://codesandbox.io/p/devbox/7ly2r6. Client is just not usable, for example I can not use `optimizelyClient.setForcedDecision()`.
![image](https://github.com/optimizely/react-sdk/assets/48070343/a5fdae7d-bc83-4b34-b32d-cd884b49eb89)


I am coming from [this issue](https://github.com/optimizely/react-sdk/issues/192), I was looking forward to the improvements, so hopefully you will manage to resolve it soon🤞

ricky-harrys:
I'm seeing this issue as well with the same integration as @dominikjasek's codesandbox

mikechu-optimizely:
[...]

PR #255, opened 2024-03-22
[FSSDK-9984] fix: initialization and setUser errors

mikechu-optimizely:
## Summary
- Addressing #235 and #236 
- Fix and add associated unit tests
- Update the underlying JS SDK to v5.3.0
- Add VS Code test config files
- Make this PR as long as humanly possible.

## Test plan

- Existing unit and integration tests are expected to pass

## Issues
- FSSDK-9984 & FSSDK-9985

### DIFF
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022-2023, Optimizely
+ * Copyright 2022-2024, Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -43,7 +43,7 @@ interface OptimizelyProviderState {
 export class OptimizelyProvider extends React.Component<OptimizelyProviderProps, OptimizelyProviderState> {
   constructor(props: OptimizelyProviderProps) {
     super(props);
-    
+
     this.setUserInOptimizely();
   }
 
@@ -82,7 +82,6 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
     // if user is a promise, setUser occurs in the then block above
     if (finalUser) {
       try {
-        //await optimizely.onReady();
         await optimizely.setUser(finalUser);
       } catch {
         logger.error('Error while trying to set user.');
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022-2024, Optimizely
+ * Copyright 2022-2023, Optimizely
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -43,9 +43,7 @@ interface OptimizelyProviderState {
 export class OptimizelyProvider extends React.Component<OptimizelyProviderProps, OptimizelyProviderState> {
   constructor(props: OptimizelyProviderProps) {
     super(props);
-  }
-
-  componentDidMount(): void {
+    
     this.setUserInOptimizely();
   }
 
@@ -84,7 +82,7 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
     // if user is a promise, setUser occurs in the then block above
     if (finalUser) {
       try {
-        await optimizely.onReady();
+        //await optimizely.onReady();
         await optimizely.setUser(finalUser);
       } catch {
         logger.error('Error while trying to set user.');
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -76,7 +76,10 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
       // deprecation warning
       logger.warn('Passing userId and userAttributes as props is deprecated, please switch to using `user` prop');
     } else {
-      finalUser = DefaultUser;
+      finalUser = {
+        id: DefaultUser.id,
+        attributes: userAttributes || DefaultUser.attributes,
+      };
     }
 
     // if user is a promise, setUser occurs in the then block above
--- a/src/Provider.tsx
+++ b/src/Provider.tsx
@@ -81,6 +81,7 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
       finalUser = DefaultUser;
     }
 
+    // if user is a promise, setUser occurs in the then block above
     if (finalUser) {
       try {
         await optimizely.onReady();
--- a/src/client.ts
+++ b/src/client.ts
@@ -19,7 +19,6 @@ import { OptimizelyDecision, UserInfo, createFailedDecision, areUsersEqual } fro
 import { notifier } from './notifier';
 import { logger } from './logger';
 import { FeatureVariableValue } from '@optimizely/optimizely-sdk';
-import { ProjectConfig } from '@optimizely/optimizely-sdk/dist/core/project_config';
 
 export type VariableValuesObject = {
   [key: string]: any;
--- a/src/client.ts
+++ b/src/client.ts
@@ -60,7 +60,7 @@ export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserConte
   onReady(opts?: { timeout?: number }): Promise<any>;
   setUser(userInfo: UserInfo): Promise<void>;
   onUserUpdate(handler: OnUserUpdateHandler): DisposeFn;
-  isReady(): boolean;
+  isReactClientReady(): boolean;
   getIsReadyPromiseFulfilled(): boolean;
   getIsUsingSdkKey(): boolean;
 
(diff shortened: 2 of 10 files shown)


src/hooks.ts
```diff
@@ -159,7 +159,7 @@ function subscribeToInitialization(
             clientReady: false,
             didTimeout: true,
           });
-          res.dataReadyPromise!.then(() => {
+          res?.dataReadyPromise?.then(() => {
```
mikechu-optimizely, line 162: `res` was `undefined` putting us in the `catch`

src/client.ts
```diff
@@ -351,10 +351,11 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
   }
 
   public async setUser(userInfo: UserInfo): Promise<void> {
+    // The JS SDK will try to assign VUID if the user ID is null
     this.setCurrentUserContext(userInfo);
 
     this.user = {
-      id: userInfo.id || DefaultUser.id,
+      id: userInfo.id || this.getUserContext()?.getUserId() || DefaultUser.id,
```
raju-opti, line 358: If userInfo.id is not provided, should we re-use previous userId or it is intended that we go back to the default it that is vuid?

src/hooks.ts
```diff
@@ -143,7 +143,7 @@ function subscribeToInitialization(
             clientReady: false,
             didTimeout: false,
           });
-          res.dataReadyPromise!.then(() => {
+          res?.dataReadyPromise?.then(() => {
```
raju-opti, line 146: can we please update it to`res.dataReadyPromise?.then()`?

res cannot be null or undefined as seen from the type, otherwise the condition inside the`if` at line 129 will throw before execution reaches this point. In this case res.dataReadyPromise is undefined.

src/hooks.ts
raju-opti, line 162: similar to previous comment

raju-opti (changes requested):

src/hooks.ts
jaeopt: dataReadyPromose can be null? If so, I do not see its handling.

jaeopt (changes requested):
Clarification regarding null dataReadyPromise.

src/client.ts
mikechu-optimizely: Thanks for asking @raju-opti . If `userInfo.id` is not provided, in client-side cases, we're looking for the `vuid_385d...` to be issued by `this.setCurrentUserContext(userInfo);`.  In server-side context, yes, we're back to default id.

raju-opti: I was asking more about cases like:

`
setUser({ id: 'mike-chu' });
// do bunch of other stuff
// then
setUser({})
`
In that case the user will want to clear the userId, but this will just pull the previous userId ('mike-chu') instead will logically changes nothing.

src/hooks.ts
mikechu-optimizely: @jaeopt in client.ts L288, we assign `dataReadyPromise` but on L298 we return from the `USER_NOT_READY` without `dataReadyPromise` (`undefined`) which seems to be as designed on L42 earlier in the file.  

The optional chain `res.dataReadyPromise?.then(() => { ...` checks for this falsey scenario?  

Do see another solution?

src/client.ts
mikechu-optimizely: Would `setCurrentUserContext` from client.tsL330 be hit?  

Let's pair up maybe on Teams?

mikechu-optimizely:
> Clarification regarding null dataReadyPromise.

@jaeopt I added an inline comment to clarify, hopefully.  DM me on Teams to help me understand the covering case you're seeing.

mikechu-optimizely:
@jaeopt and @raju-opti, Can we do a trio programming session?

mikechu-optimizely:
@jaeopt & @raju-opti When's a good time to have a sync meeting on this?

mikechu-optimizely:
We discussed a better, more proper solution.  I'm closing temporarily while I complete the WIP.

mikechu-optimizely:
@jaeopt & @raju-opti .  You can review the code, but I have tests to review and update.

src/client.ts
```diff
@@ -326,13 +338,11 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       return;
     }
 
-    if (!this.userContext || (this.userContext && !areUsersEqual(userInfo, this.user))) {
```
jaeopt, line 329: Why do we remove this if-condition? It'll be useful for avoiding the unnecessary state change.

src/client.ts
```diff
@@ -343,23 +353,46 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
   }
 
   public async fetchQualifiedSegments(options?: optimizely.OptimizelySegmentOption[]): Promise<boolean> {
-    if (!this.userContext || !this.isReady()) {
+    if (this.odpExplicitlyOff) {
+      return true;
+    }
+
+    if (!this.userContext) {
```
jaeopt, line 360: Do we need to keep track of "odpOff" here in react-sdk? It'll be tracked and returned properly in JS core.

src/Provider.tsx
```diff
@@ -81,9 +79,9 @@ export class OptimizelyProvider extends React.Component<OptimizelyProviderProps,
       finalUser = DefaultUser;
```
jaeopt, line 79: Shouldn't it be like this since we have keep userAttrobutes?
```
finalUser = {
   id: null,
  attributes: userAttributes || {},
```

src/client.ts
```diff
@@ -343,23 +353,46 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
-      await this.fetchQualifiedSegments();
+    let fetchQualifedSegmentsSucceed = false;
+    if (userInfo.id === DefaultUser.id) {
+      // if user is anonymous...
+      // wait for the SDK client to be ready before
+      await this._client?.onReady();
+      // setting the user context
+      this.setCurrentUserContext(userInfo);
+
+      this.user.id = this.userContext?.getUserId() || DefaultUser.id;
```
jaeopt, line 381: Can we add comment why we update id again here?

src/client.ts
```diff
@@ -343,23 +353,46 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
+
+      fetchQualifedSegmentsSucceed = await this.fetchQualifiedSegments();
+    } else {
+      this.setCurrentUserContext(userInfo);
+
+      this.user.id = this.userContext?.getUserId() || DefaultUser.id;
+
+      await this._client?.onReady();
+      fetchQualifedSegmentsSucceed = await this.fetchQualifiedSegments();
```
jaeopt, line 390: This and the block above for "DefaultUser.id" look same.
@raju-opti can you clarify your concern about this having same async for both?

jaeopt (changes requested):
Looks good. A few changes suggested. Wondering if we can have good tests covering all different scenarios for this complicated timing dependency.

src/client.ts
mikechu-optimizely: In the calling method, we're updating the React-side `this.user` to match the user-provided `userInfo` input. 

This method is responsible for updating the underlying JS SDK userContext. Here we've been comparing the provided `userInfo` and the (now updated) React `this.user` which skips updating the underlying JS userContext.

The appropriate SRP would be to read the underlying JS `userContext` and then compare it with the user-provided `userInfo` to see if it needs updating.  

I can do that.

src/client.ts
mikechu-optimizely: The bool returned from this method, means there were no errors.  

If ODP is expressly off, this method should assume that there were no errors and return `true` regardless if the `userContext` is truthy or not.

True, the underlying JS SDK runs this check earlier during its createInstance and sets the ODP stack to `undefined`. Based on this, the [JS client returns](https://github.com/optimizely/javascript-sdk/blob/master/lib/optimizely/index.ts#L1767)  `null` which is [interpreted by the userContext as a failure](https://github.com/optimizely/javascript-sdk/blob/master/lib/optimizely_user_context/index.ts#L261) which is interpreted by the React SDK that the user is not ready.

src/Provider.tsx
mikechu-optimizely: Updating...

src/client.ts
mikechu-optimizely: Updating...

src/client.ts
mikechu-optimizely: I agree. These are very close or could be DRY-ed out.  Whether to set the underlying userContext or try to fetchQualifiedSegments, we still need to await the JS SDK's ready.

Waiting a little bit for @raju-opti but we should move forward when ready.

jaeopt (approved):
LGTM

src/client.ts
```diff
@@ -1165,6 +1166,18 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     return this._client.getOptimizelyConfig();
   }
 
+  /**
+   * Returns the ProjectConfig object from the underlying SDK
+   * @returns {ProjectConfig | null} optimizely project config
+   */
+  public getProjectConfig(): ProjectConfig | null {
```
jaeopt, line 1173: Why do we want to make it public? If we open it, it won't be easy for future change.

jaeopt (changes requested):
Need clarification about public projectConfig

src/client.ts
mikechu-optimizely: That's excellent feedback. Yes. I don't see a reason to make it public.

mikechu-optimizely: I removed the implementation and added it to the `Omit` on L57.

jaeopt (approved):
LGTM

PR #262, opened 2024-05-16
[FSSDK-10198] odp not integrated error

junaed-optimizely:
## Summary
If the project does not have ODP integration enabled, sdk should not call the fetchQualifiedSegments method.

## Test plan
Existing test should pass

## Issues
[FSSDK-10198](https://jira.sso.episerver.net/browse/FSSDK-10198)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -54,7 +54,8 @@ export const DefaultUser: UserInfo = {
   attributes: {},
 };
 
-export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserContext' | 'getProjectConfig'> {
+export interface ReactSDKClient
+  extends Omit<optimizely.Client, 'createUserContext' | 'getProjectConfig' | 'isOdpIntegrated'> {
   user: UserInfo;
 
   onReady(opts?: { timeout?: number }): Promise<any>;
@@ -368,7 +369,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       return false;
     }
 
-    if (this.odpExplicitlyOff || !this._client?.getProjectConfig()?.odpIntegrationConfig?.integrated) {
+    if (this.odpExplicitlyOff || !this._client?.isOdpIntegrated()) {
       return true;
     }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -364,12 +364,12 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
   }
 
   public async fetchQualifiedSegments(options?: optimizely.OptimizelySegmentOption[]): Promise<boolean> {
-    if (this.odpExplicitlyOff || !this._client?.getProjectConfig()?.odpIntegrationConfig?.integrated) {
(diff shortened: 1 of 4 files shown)


mikechu-optimizely (approved):
LGTM.

kimyu92:
Looks like this https://github.com/optimizely/react-sdk/pull/262 would resolve the issue

src/client.ts
```diff
@@ -364,14 +364,14 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
...
   public async fetchQualifiedSegments(options?: optimizely.OptimizelySegmentOption[]): Promise<boolean> {
-    if (this.odpExplicitlyOff) {
-      return true;
-    }
-
     if (!this.userContext) {
       return false;
     }
 
+    if (this.odpExplicitlyOff || !this._client?.getProjectConfig()?.odpIntegrationConfig?.integrated) {
```
jaeopt, line 371: Looks good. "getProjectConfig" is not supposed to be public API. Any way to use OptimizelyConfig, which is public access for the project config.

junaed-optimizely: OptimizelyConfig does not contain odp integration information. However it contains the datafile string. Do you recommend parsing that and extract integration info from there? 

Its also worth mentioning that JS SDK has a private `isOdpIntegrated` method of `Optimizely` base class. You can give it a look here - [isOdpIntegrated](https://github.com/optimizely/javascript-sdk/blob/2f6dc05c6aa7aacc8f6e2e99c1da128fc585742a/lib/optimizely/index.ts#L1726)

Do you recommend making it public and consume that from react ?

jaeopt: I see it will be better to open optimizely.isOdpIntegrated rather than from getProjectConfig.

junaed-optimizely: Ok. This has to wait for the JS change then.

yarn.lock
mikechu-optimizely, line 1: I feel like the package.json should have also been updated.  🤔

mikechu-optimizely (changes requested):
Please review the dependencies.

yarn.lock
junaed-optimizely: The caret(^) in package.json will make sure to update it to the latest minor or patch version based on yarn.lock record.

src/client.ts
```diff
@@ -54,7 +54,8 @@ export const DefaultUser: UserInfo = {
   attributes: {},
 };
 
-export interface ReactSDKClient extends Omit<optimizely.Client, 'createUserContext' | 'getProjectConfig'> {
+export interface ReactSDKClient
```
mikechu-optimizely, line 57: Was this the linter moving this to the next line?  

Actually, I think it's probably ok to expose `isOdpIntegrated` to React SDK.

mikechu-optimizely (approved):
LGTM.

Please check your specific error is corrected.  You might also note that in the PR description as part of manual acceptance testing.

Issue #196, opened 2023-04-19
A `useEffect`'s within `useDecision` will flap the auto update listeners on every render.


epatey:
[This](https://github.com/optimizely/react-sdk/blob/da4627c1e45f2a1f92861468f06828a8990e288d/src/hooks.ts#L467) `useEffect` will mount and execute on every single render. The root cause is that `getCurrentDecision` in `useDecision` is not wrapped in `useCallback` like it is in the other hooks.

mikechu-optimizely:
Thanks for reporting this.  I've created a work ticket (FSSDK-9624) to review.

junaed-optimizely:
This has been addressed already in #273 , and a patch release [v3.2.1](https://github.com/optimizely/react-sdk/releases/tag/3.2.1) is also out with the changes. Closing..

PR #273, opened 2024-08-14
[FSSDK-10544] Refactor Hooks code + tests

junaed-optimizely:
## Summary
- SDK hooks do not follow the "Rules of Hooks".  Specially the one that states **"Do not call Hooks after a conditional return statement"**. Check the - [Official React Doc](https://react.dev/reference/rules/rules-of-hooks) for more detail
- Broken test cases fixed
- `clientReady` is set to true even though `dataReadyPromise` returns `success == false`, bug fixed
- #196 has been addressed

## Test plan
- Hook test suite is modified to fix some broken mocks and tests
- Manual testing

## Issues
- [FSSDK-10544](https://jira.sso.episerver.net/browse/FSSDK-10544)

### DIFF
--- a/src/hooks.spec.tsx
+++ b/src/hooks.spec.tsx
@@ -22,7 +22,7 @@ import { render, renderHook, screen, waitFor } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
 import { OptimizelyProvider } from './Provider';
-import { OnReadyResult, ReactSDKClient, VariableValuesObject } from './client';
+import { NotReadyReason, OnReadyResult, ReactSDKClient, VariableValuesObject } from './client';
 import { useExperiment, useFeature, useDecision, useTrackEvent, hooksLogger } from './hooks';
 import { OptimizelyDecision } from './utils';
 const defaultDecision: OptimizelyDecision = {
@@ -62,7 +62,7 @@ const mockFeatureVariables: VariableValuesObject = {
   foo: 'bar',
 };
 
-describe.skip('hooks', () => {
+describe('hooks', () => {
   let activateMock: jest.Mock;
   let featureVariables: VariableValuesObject;
   let getOnReadyPromise: any;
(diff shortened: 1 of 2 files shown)


src/hooks.ts
```diff
@@ -544,12 +591,13 @@ export const useTrackEvent: UseTrackEvent = () => {
...
     //    was provided as a promise and we need to subscribe and wait for user to become available.
-    if (optimizely.getIsUsingSdkKey() || !isClientReady) {
+    if (optimizely && (optimizely.getIsUsingSdkKey() || !isClientReady)) {
       subscribeToInitialization(optimizely, timeout, (initState) => {
         setState(initState);
       });
     }
-  }, []);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [optimizely, timeout]);
```
mikechu-optimizely, line 600: I'm curious: why not include `isClientReady` here and above in the other hooks?

src/hooks.ts
```diff
@@ -293,18 +326,25 @@ export const useExperiment: UseExperiment = (experimentKey, options = {}, overri
...
           ...prevState,
           ...getCurrentDecision(),
         }));
       }),
     [getCurrentDecision, optimizely]
   );
+
+  if (!optimizely) {
+    hooksLogger.error(
+      `Unable to use experiment ${experimentKey}. optimizely prop must be supplied via a parent <OptimizelyProvider>`
```
mikechu-optimizely, line 344: NIT: Can we move ``` The `optimizely` prop must be supplied via a parent <OptimizelyProvider>.``` as a `const` in this file?

src/hooks.ts
```diff
@@ -404,35 +449,33 @@ export const useFeature: UseFeature = (featureKey, options = {}, overrides = {})
...
   const { optimizely, isServerSide, timeout } = useContext(OptimizelyContext);
 
-  if (!optimizely) {
-    hooksLogger.error(
-      `Unable to use decision ${flagKey}. optimizely prop must be supplied via a parent <OptimizelyProvider>`
-    );
-    return [
+  const overrideAttrs = useCompareAttrsMemoize(overrides.overrideAttributes);
+
+  const defaultDecision = useMemo(
```
mikechu-optimizely, line 454: Nice fix. 👍

src/hooks.spec.tsx
```diff
@@ -176,52 +193,47 @@ describe('hooks', () => {
...
 
     it('should respect the timeout option passed', async () => {
       activateMock.mockReturnValue(null);
+      mockDelay = 100;
       readySuccess = false;
 
       render(
         <OptimizelyProvider optimizely={optimizelyMock}>
-          <MyExperimentComponent options={{ timeout: mockDelay }} />
+          <MyExperimentComponent options={{ timeout: mockDelay - 10 }} />
```
mikechu-optimizely, line 222: NIT: What do you think if we set the `mockDelay` to 0?  Is it effectively the same? 🤔

src/hooks.spec.tsx
```diff
@@ -272,7 +278,6 @@ describe('hooks', () => {
 
       //   // TODO - Wrap this with async act() once we upgrade to React 16.9
       //   // See https://github.com/facebook/react/issues/15379
```
mikechu-optimizely, line 280: I feel we should clean up comments. 

It looks like `act` is deprecated, so we'll need a new work ticket soon to clean up our uses of it.

src/hooks.spec.tsx
```diff
@@ -447,44 +451,33 @@ describe('hooks', () => {
...
-      activateMock.mockReturnValue('12345');
-      // Wait for completion of dataReadyPromise
-      await optimizelyMock.onReady().then((res) => res.dataReadyPromise);
-      await waitFor(() => expect(screen.getByTestId('result')).toHaveTextContent('true|{"foo":"bar"}|true|true')); // when clientReady
+      // When timeout is reached, but dataReadyPromise is resolved later with the feature enabled
+      await waitFor(() => expect(screen.getByTestId('result')).toHaveTextContent('true|{"foo":"bar"}|true|true'));
     });
 
     it('should gracefully handle the client promise rejecting after timeout', async () => {
-      jest.useFakeTimers();
```
mikechu-optimizely, line 486: Cool. Tests still pass. Are Jest's fake timers 👎 nowadays?

mikechu-optimizely (approved):
LGTM. 

No show-stoppers IMHO, but I have some curiosity questions 👇 DM me if it's easier.

src/hooks.ts
junaed-optimizely: Thats a great question. If client gets ready, `isClientReady` becomes `true` and triggers another execution of the `subscribeToInitialization` which is totally unnecessary in my opinion, because if client gets ready, there is no need to subscribe to anything after that. 

On the contrary, I could include the dep in the dependency array and that will cause one extra render, which is not terrible for sure. 

This is one of those rare cases where I feel like I can ignore the principles of "React".  But do let me know if you feel strongly against this. I don't mind changing it.

src/hooks.ts
junaed-optimizely: Sure! Great suggestion!

src/hooks.spec.tsx
junaed-optimizely: The value of `mockDelay` matters here. The "Client Mock" will be ready after this amount of time (miliseconds). Setting timeout less than that will trigger timeout. There is no specific logic to pick 100 for `mockDelay`. Its just a good number 😄

src/hooks.spec.tsx
junaed-optimizely: No. They are same as before. Its just these tests do not require this functionality anymore. I do not know why its here for in the first place. It might have been useful in the past, but for current suite its not required anymore.

src/hooks.spec.tsx
junaed-optimizely: I have decreased the value of milliseconds from 100 to 20 just to make the test suite a little faster

src/hooks.ts
mikechu-optimizely: Thanks for the teaching.

Issue #256, opened 2024-04-05
[BUG] setForcedDecision does not work after update to version 3.0.1


haivle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

@optimizely/react-sdk: 3.0.1

### Current Behavior

After calling `setForcedDecision`, `useDecision(name, { autoUpdate: true })` is updated.

### Expected Behavior

`useDecision` returns the latest data

### Steps To Reproduce

1. Initialise Optimizely instance
```ts
optimizelyInstance = createInstance({
        sdkKey: env.OPTIMIZELY_SDK_KEY,
        odpOptions: {
               disabled: true,
        },
        logLevel: env.IS_VERBOSE ? 1 : 4,
});

optimizelyInstance.onUserUpdate((a) => {
      optimizely.setForcedDecision({ flagKey: "flag1" }, { variationKey: "variation1" });
});
```

2. In React
```tsx
const [decision] = useDecision("flag1", { 
    autoUpdate: true 
}); 
```

### Severity

Blocking development

mikechu-optimizely:
Let me loop back to this.  I'm working to finalise #255

byron-ino:
Hey there! Was going to write up an issue about this same topic. Currently on `v3.1.0` and was running into issues forcing a user into a variation and wasn't sure if this is related to the same bug or not, but thought I would post it here.

We are currently writing a light wrapper around the `useDecision` react hook so that we do things like pass overrides and also have logic to force decisions before calling the `useDecision` hook.

Here is a simplified version of what our wrapper looks like:

```
function useDecisionWrapper(featureKey, options, overrides){ 
  const { optimizely } = useContext(OptimizelyContext);

  const [decisionContext, forcedDecision] = getForcedDecisionOverrides(featureKey); // grabs data from cookies to force a decision
  if(decisionContext && forcedDecision){
      const user = optimizely.getUserContext();
      user?.setForcedDecision(decisionContext, forcedDecision);
  }

  return useDecision(featureKey, options, overrides);
}
```

I'd assume that since we called `setForcedDecision` for the current user, calling `useDecision` would respect the forced decision but it seems to be not. Would love some feedback or suggestions if I am doing something wrong!  I will also test `v3.1.0` tomorrow! Thank you :)

mikechu-optimizely:
[...]

PR #274, opened 2024-08-15
[FSSDK-10120] setForcedDecision does not  reflects in Optimizely client instance

junaed-optimizely:
## Summary
- #256 has been addressed
- user context reset for the same user bug fix for decision methods
- Performance improvment for user prop check

## Test plan
- As we are mocking the inner client, there is no way to add test cases for this specific use case right now. However there is a dedicated ticket [FSSDK-10439](https://jira.sso.episerver.net/browse/FSSDK-10439) to improve the "Optimizely client" test coverage. Improvement should be expected from there.

## Issues
[FSSDK-10120](https://jira.sso.episerver.net/browse/FSSDK-10120)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -361,6 +361,12 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
       return null;
     }
 
+    if (this.userContext && areUsersEqual(userInfo, this.user)) {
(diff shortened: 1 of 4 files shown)


mikechu-optimizely (approved):
LGTM as our per our collab DMs

Issue #265, opened 2024-06-19
[BUG] Can't set log level


AntonRatnick:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

3.1.1

### Current Behavior

Setting log level to `Error`, don't reduce amount of warnings and info logs in console

### Expected Behavior

If log `level == Error` no info or warning logs appear in console.

### Steps To Reproduce

1. [Changing log level to Error, the way described in the documentation not changing console output](https://docs.developers.optimizely.com/feature-experimentation/docs/customize-logger-react)
2. Passing log level as part of config to `createInstance` also not working

### Link

1. https://github.com/optimizely/react-sdk/blob/master/src/logger.tsx#L19 Here we create logger instance and pass only prefix.
2. https://github.com/optimizely/javascript-sdk/blob/4909efb74335b754592ddf6e8415047fc0bf3e7b/lib/modules/logging/logger.ts#L99 
If I am not mistaken this is the logger we creating, without any arguments beside the prefix.
3. https://github.com/optimizely/react-sdk/blob/d43cabd6f0181d504951d74822a0efd4f28aec7c/src/client.ts#L20 
Here we import created logger, and use it.
4. The logger instance created for React SDK is not exported, so changing log level on created instance also not possible.

alexparish:
I was just in the process of creating a bug for this exact scenario. Luckily saw this one first!

It's frustrating that, as a consumer, I cannot change the log level of the react-sdk internal logger as it is causing a lot of unnecessary log noise for my services in edge case scenarios.

Bundas:
I was poking around and it seems like the easiest solution is to export the instance of logger in `src/logger.tsx` called `logHandler`. Once it is exported, we can access it at a constructor of `OptimizelyReactSDKClient` in `src/client.ts` and set a proper log level based on the config there. I tested it locally and it works like a charm. Unfortunately I can't create a PR at this time, so if anybody has a capacity...

junaed-optimizely:
This definitely requires some attention. There are some inconsistencies that I can confirm. I have created an internal ticket FSSDK-10554 for further investigation.

junaed-optimizely:
@AntonRatnick, there is a patch release [v3.2.2](https://github.com/optimizely/react-sdk/releases/tag/3.2.2) out with the fix. Could you please check and confirm?

PR #276, opened 2024-08-20
[FSSDK-10554] Can't set log level error

junaed-optimizely:
## Summary
 - #265 has been addressed. 
 - Instead of creating a new log for the React client,  we can reuse the global log that `Optimizely` already provides with its `getLogger` API. That way we can configure the log system as a whole from a single source. 
 - FYI, we are already using `getLogger` for [hooks](https://github.com/optimizely/react-sdk/blob/c16ea02cb423c78b9de09b2e78e0d40b590ee38a/src/hooks.ts#L27) and [providers](https://github.com/optimizely/react-sdk/blob/c16ea02cb423c78b9de09b2e78e0d40b590ee38a/src/Provider.tsx#L26C1-L26C50).
## Test plan
Existing test should pass

## Issues
[FSSDK-10554](https://jira.sso.episerver.net/browse/FSSDK-10554)

### DIFF
--- a/src/logger.tsx
+++ b/src/logger.tsx
@@ -16,7 +16,8 @@
 
 import * as optimizely from '@optimizely/optimizely-sdk';
 import { sprintf } from './utils';
-const logHandler = optimizely.logging.createLogger({ prefix: '[React-SDK]' });
+
+const logHandler = optimizely.getLogger('ReactSDK');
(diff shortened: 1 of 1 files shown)


mikechu-optimizely (approved):
LGTM

PR #278, opened 2024-09-02
[FSSDK-10439] client provider test

junaed-optimizely:
## Summary
Significant improvements were made to the test coverage of the "Client" and "Provider files.
- Provider Coverage: Increased from 72.34% to 100%
- Client Coverage: Improved from 81.4% to 99.27% by restructuring and adding tests for various methods
- Test Additions: A total of 52 new tests were added, covering previously missing cases and deprecated API updates.

The coverage reports have been generated with the help of "Istanbul" that is the default coverage tool of "Jest". Here are the commands that I have used to generate the coverage report - 

```bash
# for client
npx jest --coverage -- client.spec.ts
# and for provider
npx jest --coverage -- provider.spec.tsx
```

## Test plan
A total of 52 new tests were added, covering previously missing cases and deprecated API updates.

## Issues
- [FSSDK-10439](https://jira.sso.episerver.net/browse/FSSDK-10439)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ build/
 
 # test artifacts
 **.tgz
+coverage/
--- a/src/Provider.spec.tsx
+++ b/src/Provider.spec.tsx
@@ -16,18 +16,13 @@
 
 /// <reference types="jest" />
 
-//jest.mock('./client');
-
 import React from 'react';
-import { render, act } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import { OptimizelyProvider } from './Provider';
-import { DefaultUser, ReactSDKClient, createInstance } from './client';
+import { DefaultUser, ReactSDKClient } from './client';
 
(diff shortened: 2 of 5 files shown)


mikechu-optimizely (approved):
What are your thoughts on splitting out client.spec.ts into separate files? Along with less cognitive load for review/maintenance, I think we get some speed of execution benefit too.  🤔 

Otherwise,  LGTM

junaed-optimizely:
> NIT: I like a bit of whitespace between test blocks, and even to separate the arrange-act-assert within each test. imho

Thats a good suggestion. Let me do that

junaed-optimizely:
> What are your thoughts on splitting out client.spec.ts into separate files? Along with less cognitive load for review/maintenance, I think we get some speed of execution benefit too. 🤔
> 
> Otherwise, LGTM

I like your suggestion. But its not uncommon to have a single test file against a single module. Its offers easier context management as well.

PR #284, opened 2024-09-25
[FSSDK-10665] fix: Github Actions YAML files vulnerable to script injections corrected

FarhanAnjum-opti:
## Summary
Instead of directly using variables that can contain user input, new intermediate variables were introduced so that script injection attacks could be mitigated.

## Issues
- [FSSDK-10665](https://jira.sso.episerver.net/browse/FSSDK-10665)


junaed-optimizely (approved):
LGTM! Thanks!

PR #285, opened 2024-10-02
[FSSDK-10730] isClientReady logic adjustment

junaed-optimizely:
## Summary
React SDK sets client.isClientReady property to true immediately if datafile presents in the config, before projectConfigManager gets ready, which should not be the case. 

## Test plan
Existing test should pass

## Issues
[FSSDK-10730](https://jira.sso.episerver.net/browse/FSSDK-10730)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -240,7 +240,6 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     });
 
     this._client = optimizely.createInstance(configWithClientInfo);
-    this.isClientReady = !!configWithClientInfo.datafile;
     this.isUsingSdkKey = !!configWithClientInfo.sdkKey;
 
     if (this._client) {
@@ -251,7 +250,6 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
           this.isClientReady = clientResult.success;
           this.isUserReady = userResult.success;
           const clientAndUserReady = this.isClientReady && this.isUserReady;
-
           this.clientAndUserReadyPromiseFulfilled = true;
 
           return {
--- a/src/client.ts
+++ b/src/client.ts
@@ -240,6 +240,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     });
 
     this._client = optimizely.createInstance(configWithClientInfo);
+    this.isClientReady = !!this.getOptimizelyConfig();
     this.isUsingSdkKey = !!configWithClientInfo.sdkKey;
 
     if (this._client) {
@@ -249,7 +250,7 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
         ([userResult, clientResult]) => {
           this.isClientReady = clientResult.success;
           this.isUserReady = userResult.success;
-          const clientAndUserReady = this.isClientReady && this.isUserReady;
+          const clientAndUserReady = this.isReady();
           this.clientAndUserReadyPromiseFulfilled = true;
 
           return {


src/client.ts
```diff
@@ -240,7 +240,6 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
     });
 
     this._client = optimizely.createInstance(configWithClientInfo);
-    this.isClientReady = !!configWithClientInfo.datafile;
```
raju-opti, line 243: I guess this was done to enable synchronous rendering to facilitate SSR. If this is removed, does server side rendering provide correct decision?

junaed-optimizely: Its possible to write asynchronous code in a SSR framework. But it must be handled before the component is rendered. Framework like `Next` has this `getServerSideProps` or "Server Components" to handle this scenario.

So we can resolve `onReady` promise before the component is rendered. Thats an option.

Our SDK doc around [SSR](https://github.com/optimizely/react-sdk/tree/master?tab=readme-ov-file#server-side-rendering) seems little confusing to me. 

Now to your point, I am assuming the static check is here for SSR. But how do we make sure that the datafile is valid?

junaed-optimizely: Also isn't it a bug in CSR ? It could set the clientReady to true immediately and then later it could be false!!

junaed-optimizely: **More context on this.**

We are passing `isServerSide` props in the `OptimizelyProvider` and it has no use in the React Client. We are only using it in hooks like - [useDecision](https://github.com/optimizely/react-sdk/blob/1197fb2483aeb3ba6c85c18d9e605663da73304f/src/hooks.ts#L468). I guess you are talking about this. Here it is checking `isServerSide` and setting clientReady to true immediately. But I guess it still relies on the framework to resolve the `onReady` promise first, and if thats the case then `optimizely?.isReady()` will return proper result. I am just curious about the necessity of `isServerSide` in that case? Maybe I am missing something?

junaed-optimizely: **More on this**

Sorry I am dumping lots of things here. I am putting this here for reference. I feel like we need to discuss more on this maybe on a meeting.

[Provider.componentDidUpdate](https://github.com/optimizely/react-sdk/blob/1197fb2483aeb3ba6c85c18d9e605663da73304f/src/Provider.tsx#L94) has a check. SSR does not execute  any lifecycle methods / hooks like - `componentDidMount()`, and `componentDidUpdate()` or `useEffect()` . So this check is invalid.

raju-opti: I am not super familiar with all these things myself. As it is used now, isServerSide is probably just same as checking whether a datafile is available synchronously. 

Regarding `isClientReady`, I think what this exactly means is not precise in the current implementation. I think we should rigorously define what it means for the client to be ready in the first place. So this is an area for refactor. If this is set without checking the datafile validity, it can later become false. So yeah, the removed looks wrong. I don't have any issue removing this, I just wanted to ensure that SSR render works when we remove this. If it does, we can go ahead and remove this.

Let's discuss it further in a call.

junaed-optimizely: Cool. I will check if everything works as expected in SSR with / without it before our 1:1

src/client.ts
```diff
@@ -251,7 +250,6 @@ class OptimizelyReactSDKClient implements ReactSDKClient {
           this.isClientReady = clientResult.success;
           this.isUserReady = userResult.success;
           const clientAndUserReady = this.isClientReady && this.isUserReady;
```
mikechu-optimizely, line 252: NIT: Off topic: Do you want to use `const clientAndUserReady = this.isReady();`

mikechu-optimizely (approved):
LGTM.

I agree with the removal after reading the ticket.  Thanks for testing CSR & SSR.

src/client.ts
junaed-optimizely: Yeah. I could do that. Great find!

raju-opti (approved):
lgtm

PR #287, opened 2024-11-06
[FSSDK-8184] override attribute bug fix of track method

junaed-optimizely:
## Summary
While checking the second parameter of `OptimizelyReactClient.track`, to decide if its a `userId` or `eventTags`, `overrideAttributes` is set to `undefined` if the second argument is `eventTags`, which should not be the case. userId and attribute overrides are two seperate things.

## Test plan
- Test structure of `track` method of React client has improved
- New test case added to cover this specific scenario

## Issues
- [FSSDK-8184](https://jira.sso.episerver.net/browse/FSSDK-8184)

### DIFF
--- a/src/client.spec.ts
+++ b/src/client.spec.ts
@@ -1423,7 +1423,7 @@ describe('ReactSDKClient', () => {
         expect(mockFn).toHaveBeenCalledWith('evt1', 'user3', { bla: 'bla' }, { tagKey: 'tagVal' });
       });
 
-      it.only('track with event key, tags, and overrided attributes, calls inner client with valid arguments', () => {
+      it('track with event key, tags, and overrided attributes, calls inner client with valid arguments', () => {
         const mockFn = mockInnerClient.track as jest.Mock;
         instance.track('evt1', { tagKey: 'tagVal' }, { bla: 'bla' });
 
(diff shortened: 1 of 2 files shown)


src/client.spec.ts
```diff
@@ -1390,36 +1390,46 @@ describe('ReactSDKClient', () => {
...
-      expect(mockFn).toHaveBeenCalledTimes(1);
-      expect(mockFn).toHaveBeenCalledWith('evt1', 'user2', { bar: 'baz' }, undefined);
+        expect(mockFn).toHaveBeenCalledTimes(1);
+        expect(mockFn).toHaveBeenCalledWith('evt1', 'user2', { bar: 'baz' }, undefined);
+      });
 
-      mockFn.mockReset();
+      it('track with event key and event tags, calls inner client with valid arguments', () => {
+        const mockFn = mockInnerClient.track as jest.Mock;
+        instance.track('evt1', { tagKey: 'tagVal' });
```
mikechu-optimizely, line 1412: Would it be good to explicitly pass `undefined` as the second parameter or use named parameters here and the last test of this `describe` block?

mikechu-optimizely (commented):
LGTM.

Do need to cover if track's `overrideUserId` is `null` or if `_client` is not initialized / `null`?  Maybe overkill?

mikechu-optimizely (approved):

src/client.spec.ts
junaed-optimizely: `_client=null` is already covered in the previous block

Issue #300, opened 2024-12-07
[BUG] optimizely.isReady() always returns false in version 3.2.3


mlazari:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK Version

3.2.3

### Current Behavior

If I use a OptimizelyProvider that on first render has user undefined, then the `isReady()` method of the optimizely client will always return false, even if the user is set later.
It seems to be related to this change that added verification for isUserReady to isReady(): https://github.com/optimizely/react-sdk/commit/62a12ff9bd1180335135e77124d27b6f0966ca1d#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R447

`isUserReady` seems to be changed only once here - https://github.com/optimizely/react-sdk/blob/3.2.3/src/client.ts#L252 - after `userReadyPromise` is resolved here when `setUser` is called first time (e.g. from OptimizelyProvider) - https://github.com/optimizely/react-sdk/blob/3.2.3/src/client.ts#L410. If `setUser` was called with the default user (with id = null), then the promise gets resolved with `{ success: false }`, thus `isUserReady` will remain false and never change again.

### Expected Behavior

`isReady()` should return true at some point when the optimizely client is ready.

### Steps To Reproduce

1. Clone https://github.com/mlazari/OptimizelyRepro and make sure you are in the main branch
2. Copy the `.env.development` file to a new `.env.local` file in the root of the project and set your optimizely key in it
3. Run `npm install`
4. Run `npx expo prebuild --clean`
5. Run the app with `npx expo run:ios` (or `npx expo run:android`)
6. Tap on the "Re-check" button, see the "Optimizely Ready" field remains false. Try again by tapping on "Set user info", then again on "Re-check", see it still says "false"

https://github.com/user-attachments/assets/7f90556e-03eb-4d5e-8294-edc331dd7aa2

7. Close the app and metro bundler
8. Switch to branch 3.2.2 (which has `@optimizely/react-sdk` downgraded to 3.2.2)
9. Run `npm install`
10. Run `npx expo prebuild --clean`
11. Run the app with `npx expo run:ios` (or `npx expo run:android`)
12. Tap on the "Re-check" button, see the "Optimizely Ready" field changes to true

https://github.com/user-attachments/assets/ed5939b3-2a56-40e7-bf02-10af50c1cbb9

### Link

https://github.com/mlazari/OptimizelyRepro/blob/main/app/index.tsx

[...]

PR #302, opened 2024-12-24
Junaed/fssdk 10980 optimizely onready always returns false

junaed-optimizely:
## Summary
`optimizelyClient.isReady` always returns `false`  when ODP is explicitly off and user id is null.

Expected behavior - "User promise should not be fulfilled unless a valid user is set, when ODP is disabled".

## Test plan
Test case has been added to cover this edge case

## Issues
- #300 
- [FSSDK-10980](https://jira.sso.episerver.net/browse/FSSDK-10980)


raju-opti (approved):
LGTM

PR #329, opened 2026-04-13
[FSSDK-12454] cleanup

junaed-optimizely:
## Summary
- Remove all legacy top-level source files from src/ (24 files) that were superseded by the new subdirectory-based modules (`client/`, `hooks/`, `logger/`, `provider/`, `utils/`)
- Remove Jest and related packages (`jest, jest-environment-jsdom, ts-jest, @types/jest`), config (`jest.config.js`), and the `test:legacy` script - all tests now run exclusively via `Vitest`
- Update all internal import paths to remove explicit `/index` suffixes, preventing accidental resolution to old top-level files
- Simplify `Vitest` config by removing the per-subdirectory include list (no longer needed since old spec files are gone)
- Re-enable `lint-staged` with `eslint --fix` (was temporarily disabled during v6 migration)
- Remove Jest-specific `.vscode/launch.json` and `jest: true` from `.eslintrc.js` env

## Test plan
- All 249 tests pass across 14 test files via vitest run
## Issues
- FSSDK-12454


src/server.ts
```diff
@@ -31,6 +30,13 @@ export {
   createOdpManager,
   createVuidManager,
   createErrorNotifier,
+  OptimizelyDecideOption,
+  LogLevel,
+  NOTIFICATION_TYPES,
+  DECISION_NOTIFICATION_TYPES,
+  DECISION_SOURCES,
+  eventDispatcher,
+  getSendBeaconEventDispatcher,
```
raju-opti: we should not need this in server components

raju-opti (approved):

PR #332, opened 2026-05-04
[FSSDK-12471] provider fix

junaed-optimizely:
## Summary
- Remove the cleanup effect that called store.reset() and UCM.dispose() on unmount
- Merge the onReady and CONFIG_UPDATE effects into a single effect with a configReceived flag
### StrictMode compatibility (React 18+):
In dev mode, StrictMode runs effect cleanup→remount without re-running the render body. The cleanup effect called store.reset() (clearing userContext to null) and UCM.dispose() (aborting in-flight async operations like ODP segment fetches). Since the Provider doesn't re-render after effect remount, nobody restores the state — hooks stay stuck in isLoading: true forever. On real unmount, the store is garbage collected with the component tree, so eager reset is unnecessary.

### Cached datafile race condition:
When the datafile response is served from the browser HTTP cache, the polling config manager loads it before the effect subscribes its CONFIG_UPDATE listener. Since the config doesn't change after subscription, the notification is never re-emitted, and hooks never transition from hasConfig: false to true. onReady().then() acts as a fallback for this case; the configReceived flag prevents a double-refresh when both fire on a cold cache.

## Test plan
- Updated existing Provider specs to reflect new unmount behavior (store state preserved, not eagerly reset)
- Manually verified in a Next.js 15 + React 19  and React 17 app
## Issues
FSSDK-12471

### DIFF
--- a/src/provider/OptimizelyProvider.tsx
+++ b/src/provider/OptimizelyProvider.tsx
@@ -70,8 +70,8 @@ export function OptimizelyProvider({
     userManagerRef.current.resolveUserContext(user, qualifiedSegments, skipSegments);
   }
 
-  // Effect: Client onReady — only needed for error handling.
-  // Readiness is derived from userContext + getOptimizelyConfig() by hooks.
+  // Effect: Client readiness + config update subscription.
+  // Handles both initial datafile fetch and subsequent polling updates.
   useEffect(() => {
     if (!client) {
       console.error('[OPTIMIZELY - REACT] OptimizelyProvider must be passed an Optimizely client instance');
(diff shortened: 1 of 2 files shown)


src/provider/OptimizelyProvider.tsx
```diff
@@ -80,42 +81,38 @@ export function OptimizelyProvider({
...
     );
 
-    return () => {
-      client.notificationCenter.removeNotificationListener(listenerId);
-    };
-  }, [client, store]);
+    client
+      .onReady({ timeout })
+      .then(() => {
+        if (!isMounted || configReceived || hadConfigAtRender) return;
```
raju-opti: there is a corner case here:
1. let's say at initial render, there is a config (v1)
2. before this effect runs, there comes a config update(v2). This v2 will be missed by the notification listener above as well cause it came before the effect ran.

the onReady.then here will see that hadConfigAtRender = true and will not refresh either.


one way to deal with this would be:
```js
//in component body
const cofigVersionAtRender = optimizely.getOptimizelyConfig().revision

client.onReady().then(() => {
  if (!isMounted) return;

  const currentConfigVersion =  optimizely.getOptimizelyConfig().revision
  if (currentConfigVersion != cofigVersionAtRender) {
   store.refresh()
  }

  // subscribe notificationListener her for all subsequent updates
  optimizely.notificationCenter.notificationCenter.addNotificationListener(
      NOTIFICATION_TYPES.OPTIMIZELY_CONFIG_UPDATE,
  )
})
```

we can also get rid of the `configReceived` variable

raju-opti (changes requested):

raju-opti (changes requested):
plese see comment

src/provider/OptimizelyProvider.tsx
junaed-optimizely: Good find. Updated!

raju-opti (approved):

PR #336, opened 2026-05-15
[FSSDK-12647] user id vuid adjustment

junaed-optimizely:
## Summary
- Add null support to the Provider's user prop type
- VUID mode now requires an explicit empty object (user={{}}) instead of omitting user
## Test plan
Tests have been added to support the change

## Issues
- FSSDK-12647

### DIFF
--- a/src/provider/OptimizelyProvider.spec.tsx
+++ b/src/provider/OptimizelyProvider.spec.tsx
@@ -233,7 +233,7 @@ describe('OptimizelyProvider', () => {
       let capturedContext: OptimizelyContextValue | null = null;
 
       const { unmount } = render(
-        <OptimizelyProvider client={mockClient}>
+        <OptimizelyProvider client={mockClient} user={{ id: 'user-1' }}>
           <ContextConsumer onContext={(ctx) => (capturedContext = ctx)} />
         </OptimizelyProvider>
       );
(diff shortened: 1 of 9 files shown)


src/provider/OptimizelyProvider.spec.tsx
```diff
@@ -793,6 +793,56 @@ describe('OptimizelyProvider', () => {
...
+      expect(mockClient.createUserContext).not.toHaveBeenCalled();
+      expect(capturedContext!.store.getState().userContext).toBeNull();
+
+      rerender(
+        <OptimizelyProvider client={mockClient} user={{ id: 'user-1' }}>
+          <ContextConsumer onContext={(ctx) => (capturedContext = ctx)} />
+        </OptimizelyProvider>
+      );
+
+      expect(mockClient.createUserContext).toHaveBeenCalledWith('user-1', undefined);
```
raju-opti, line 842: can we also add an assertion that the userContext is present in the store?

src/provider/OptimizelyProvider.spec.tsx
```diff
@@ -793,6 +793,56 @@ describe('OptimizelyProvider', () => {
...
+    });
+  });
+
```
raju-opti, line 867: can we also add a test for when user prop goes from valid to null that the store userCOntext also set to null?

raju-opti (approved):

